PR #1, opened 2021-06-02
Add RandomVariable log-likelihoods

brandonwillard:
This PR adds an initial log-likelihood entry-point function and log-pdf/pmf functions for most of the `aesara.tensor.random.basic` `RandomVariable`s.

The general idea is that the `logpdf` dispatch function simply adds `logpdf`s in the spirit of `scipy.stats`.  These implementations should be flexible when it comes to the shapes of the observed/value variables (e.g. the `RandomVariable` shouldn't need to be resized to produce a graph).

The `loglik` functions&mdash;and the associated `_loglik` dispatch functions&mdash;compute log-likelihoods for graphs potentially containing multiple `RandomVariable`s.  

`loglik` needs to be entirely rewritten, though.  Simply put, we would be better off using the `Optimization` framework.  See [the comments here](https://github.com/pymc-devs/pymc3/pull/4653#pullrequestreview-641633382) for more information.  

Nevertheless, we'll probably merge this before everything is fixed, just to establish something from which we can base our log-likelihood-requiring work.

### DIFF
--- a/aeppl/loglik.py
+++ b/aeppl/loglik.py
@@ -1,0 +1,209 @@
+from collections.abc import Mapping
+from functools import singledispatch
+from typing import Dict, Optional, Union
+
+import aesara.tensor as at
+from aesara import config
+from aesara.gradient import disconnected_grad
+from aesara.graph.basic import Constant, clone, graph_inputs, io_toposort
+from aesara.graph.fg import FunctionGraph
+from aesara.graph.op import Op, compute_test_value
+from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.random.opt import local_subtensor_rv_lift
+from aesara.tensor.subtensor import (
+    AdvancedIncSubtensor,
+    AdvancedIncSubtensor1,
+    AdvancedSubtensor,
+    AdvancedSubtensor1,
+    IncSubtensor,
+    Subtensor,
+)
+from aesara.tensor.var import TensorVariable
+
+from aeppl.logpdf import _logpdf
+from aeppl.utils import (
+    extract_rv_and_value_vars,
+    indices_from_subtensor,
+    rvs_to_value_vars,
+)
+
+
+def loglik(
+    var: TensorVariable,
+    rv_values: Optional[
+        Union[TensorVariable, Dict[TensorVariable, TensorVariable]]
+    ] = None,
+    **kwargs,
+) -> TensorVariable:
+    """Create a measure-space (i.e. log-likelihood) graph for a random variable at a given point.
+
+    The input `var` determines which log-likelihood graph is used and
+    `rv_value` is that graph's input parameter.  For example, if `var` is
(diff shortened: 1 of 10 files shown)


aeppl/logpdf.py
```diff
@@ -0,0 +1,455 @@
...
+        res, at.all(at.gt(alpha, 0.0)), at.all(at.gt(beta, 0.0))
+    )
+    return res
+
+
+@_logpdf.register(arb.VonMisesRV)
+def vonmises_logpdf(op, value, *inputs, **kwargs):
+    mu, kappa = inputs[3:]
+    res = kappa * at.cos(mu - value) - at.log(2 * np.pi) - at.log(at.i0(kappa))
+    # This doesn't match `scipy.stats.vonmises.logpdf`:
```
ricardoV94, line 213: I am not sure about the logoed, but in general thenscipy vonmises is really poorly implemented, special when it comes to the circular limits and mu translations

aeppl/logpdf.py
```diff
@@ -0,0 +1,455 @@
...
+
+@_logpdf.register(arb.NormalRV)
+def normal_logpdf(op, value, *inputs, **kwargs):
+    mu, sigma = inputs[3:]
+    res = (
+        -0.5 * at.pow((value - mu) / sigma, 2)
+        - at.log(np.sqrt(2.0 * np.pi))
+        - at.log(sigma)
+    )
+    res = Assert("sigma > 0")(res, at.all(at.gt(sigma, 0.0)))
```
ricardoV94, line 71: The Assets sound neat compared to the pymc bound. It also means we might be able to disable them if we know it's safe to on the spot during optimizations, no?

aeppl/loglik.py
```diff
@@ -0,0 +1,209 @@
...
+):
+    """Create a graph for the log-likelihood of a ``Variable``.
+
+    This function dispatches on the type of ``op``.  If you want to implement
+    new graphs for an ``Op``, register a new function on this dispatcher.
+
+    The default returns a graph producing only zeros.
+
+    """
+    value_var = rvs_to_values.get(var, var)
```
ricardoV94, line 143: My gut feeling when I was working on this, is that this default is more error prone. I feel most Aesara ops in a generative model would have some impact on the total logp. Wouldn't it be safer to default to NotImplementedError and explicitly dispatch those that should indeed return zero?

aeppl/logpdf.py
brandonwillard: Yes, they can be optimized out.

aeppl/loglik.py
brandonwillard: All this is going to be refactored.  We only need the `logpdf`s for `aehmc` right now.

aeppl/logpdf.py
brandonwillard: Let's create a separate issue for this.

PR #9, opened 2021-06-16
Implement an approach to log-probability transforms

### LINKED ISSUE #7: Implement HMC-related `RandomVariable` transforms
We now have a good framework for easily implementing the HMC-related `RandomVariable` transforms (or, as the ML people have (re)labeled them, "normalizing flows").  

We can start by creating a single Aesara rewrite (e.g. using `aesara.graph.opt.local_optimizer`) that replaces transformed `RandomVariable`s with an intermediate form in the "sample-space" graph (i.e. the first input to `joint_logprob`).  That intermediate form could be a new `RandomVariable` type (e.g. `TransformedRV`) with a `_logprob` dispatch that implements the change-of-variables formula (including the Jacobian).  This new `RandomVariable` type would essentially be a wrapper around the original `RandomVariable` that only serves to provide a distinct `_logprob` dispatch.

brandonwillard:
This PR introduces an approach to log-probability transforms.

Closes #7

### DIFF
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -1,25 +1,27 @@
 import warnings
 from collections import deque
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
 from aesara import config
 from aesara.graph.basic import graph_inputs, io_toposort
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import compute_test_value
-from aesara.graph.opt_utils import optimize_graph
+from aesara.graph.opt import GlobalOptimizer, LocalOptimizer
+from aesara.graph.optdb import OptimizationQuery
 from aesara.tensor.basic_opt import ShapeFeature
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.var import TensorVariable
 
 from aeppl.logprob import _logprob
-from aeppl.opt import PreserveRVMappings, RVSinker
+from aeppl.opt import PreserveRVMappings, logprob_canonicalize
 from aeppl.utils import rvs_to_value_vars
 
 
 def joint_logprob(
     var: TensorVariable,
     rv_values: Optional[Dict[TensorVariable, TensorVariable]] = None,
-    warn_missing_rvs=True,
+    warn_missing_rvs: bool = True,
+    extra_rewrites: Optional[Union[GlobalOptimizer, LocalOptimizer]] = None,
     **kwargs,
 ) -> TensorVariable:
     r"""Create a graph representing the joint log-probability/measure of a graph.
@@ -70,6 +72,9 @@ def joint_logprob(
         When ``True``, issue a warning when a `RandomVariable` is found in
         the graph and doesn't have a corresponding value variable specified in
         `rv_values`.
+    extra_rewrites
+        Extra rewrites to be applied (e.g. reparameterizations, transforms,
+        etc.)
 
     """
     # Since we're going to clone the entire graph, we need to keep a map from
@@ -81,7 +86,7 @@ def joint_logprob(
 
     # We add `ShapeFeature` because it will get rid of references to the old
     # `RandomVariable`s that have been lifted; otherwise, it will be difficult
-    # to give good warnings when an unaccounted for `RandomVairiable` is
+    # to give good warnings when an unaccounted for `RandomVariable` is
     # encountered
     fgraph = FunctionGraph(
         outputs=[var],
@@ -101,7 +106,10 @@ def joint_logprob(
 
     fgraph.attach_feature(rv_remapper)
 
-    _ = optimize_graph(fgraph, custom_opt=RVSinker())
+    logprob_canonicalize.query(OptimizationQuery(include=["basic"])).optimize(fgraph)
+
+    if extra_rewrites is not None:
+        extra_rewrites.optimize(fgraph)
 
     # This is the updated random-to-value-vars map with the
     # lifted variables
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -138,8 +138,11 @@ def joint_logprob(
             )
 
             q_logprob_var = _logprob(
-                node.op, q_rv_value_var, *value_var_inputs, **kwargs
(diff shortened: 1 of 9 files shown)


aeppl/joint_logprob.py
```diff
@@ -81,7 +86,7 @@ def joint_logprob(
 
     # We add `ShapeFeature` because it will get rid of references to the old
     # `RandomVariable`s that have been lifted; otherwise, it will be difficult
-    # to give good warnings when an unaccounted for `RandomVairiable` is
+    # to give good warnings when an unaccounted for `Antimicrobial` is
```
twiecki, line 89: What's that?

aeppl/transforms.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @classmethod
+    def log_jac_det(cls, value, *inputs):
+        """Construct the log of the absolute value of the Jacobian determinant."""
+        phi_inv = cls.backward(value, *inputs)
+        return at.log(at.abs_(jacobian(phi_inv, [value])))
+
+
+@_logprob.register(TransformedRV)
+def trans_uniform_logprob(op, value, *inputs, **kwargs):
```
ricardoV94, line 39:
```suggestion
def trans_logprob(op, value, *inputs, **kwargs):
```

aeppl/transforms.py
```diff
@@ -0,0 +1,146 @@
...
+
+    This function dispatches on the type of ``op``, which should be a subclass
+    of ``RandomVariable``.  If you want to implement new transforms for a
+    ``RandomVariable``, register a function on this dispatcher.
+
+    """
+    return None
+
+
+class TransUniformRV(TransformedRV):
```
ricardoV94, line 114: Are we going to need a class for each transformable variable or this is just to illustrate a custom transform case?

Can't we have an `interval transform` and dispatch it to a lists of appropriate RandomVariable types?

brandonwillard: Yes, this is just an illustration, and we can always generalize when it's useful.

aeppl/joint_logprob.py
brandonwillard: Ha, must've been in my kill ring; no idea how that word would've gotten there, though.

brandonwillard:
All right, it should be working now.

aeppl/transforms.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if name:
+        logprob.name = f"{name}_logprob"
+        jacobian.name = f"{name}_logprob_jac"
+
+    return logprob + jacobian
+
+
+@local_optimizer(tracks=None)
+def transform_logprob(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]:
```
ricardoV94, line 54: Would transform_rv be a better name?

brandonwillard: Yeah

brandonwillard:
I'll start adding the other transforms and set up a more general testing framework (based on the current one).

aeppl/transforms.py
```diff
@@ -0,0 +1,200 @@
...
+    if not isinstance(node.op, RandomVariable):
+        return
+
+    trans_node = _transformed_rv(node.op, node)
+
+    if trans_node is not None:
+        # Get the old value variable and remove it from our value variables map
+        rv_var = node.outputs[1]
+        old_value_var = rv_map_feature.rv_values[rv_var]
+        del rv_map_feature.rv_values[rv_var]
```
ricardoV94, line 101:
```suggestion
        old_value_var = rv_map_feature.rv_values.pop(rv_var)
```

aeppl/transforms.py
```diff
@@ -0,0 +1,200 @@
...
+def create_transformed_rv_op(rv_op: Op, transform: RVTransform, cls_dict_extra=None):
+
+    trans_name = getattr(transform, "name", "")
+    rv_type_name = type(rv_op).__name__
+    cls_dict = type(rv_op).__dict__.copy()
+    rv_name = cls_dict.get("name", "")
+    if rv_name:
+        cls_dict["name"] = (f"{rv_name}_{trans_name}",)
+    cls_dict = {
+        "name": f"uniform_{trans_name}",
```
ricardoV94, line 180: `uniform` shouldn't be in there anymore

brandonwillard:
I need to look into this Dirichlet transform; it would be better if we didn't have to drop a dimension.  At the very least, it complicates numerically producing its Jacobian determinant.

ricardoV94:
This logic should be easy to apply (kind of in reverse) when someone asks for the logp graph of `y = np.exp(at.Normal(0, 1))`

aeppl/utils.py
```diff
@@ -32,6 +32,13 @@
 ]
 
 
+def logsumexp(x, axis=None, keepdims=True):
```
ricardoV94, line 35: This graph will be "double" optimized, unless you place the switch inside the sum. You can also leave the max trick out since it's a default aesara rewrite now.

brandonwillard: If it's rewriting that log-sum, then we probably need to make sure that doesn't cause problems&mdash;and do something about it, if it does.

ricardoV94: It doesn't cause problems, it just adds an unecessary extra switch because the returned graph still contains a `log(sum(exp(x)))` structure even if `x` is now `x - switch(isinf(max(x), 0, max(x))`

ricardoV94: You can either let aesara do the rewrite or explicitly return the stable form `max(x) + log(sum(switch(isinf(max(x)), np.exp(x), np.exp(x-max(x))`

ricardoV94: Maybe we should add a helper `logsumexp` to aesara? To avoid reinventing it everywhere

brandonwillard: Definitely

brandonwillard:
We should have all the basic transforms and their tests in place now.

ricardoV94 (approved):
Looks good to me. I won't have time to actual run some code until tomorrow.

One example that I find useful for testing transforms is the uniform with random parameters.

PR #19, opened 2021-06-22
Implement a general mixture log-probability rewrite

brandonwillard:
This PR implements log-probabilities for general mixtures (e.g. heterogeneous `RandomVariable`s).

This implementation does not produce a **marginalized** log-probability.

It covers cases like the following:
```python
import aesara.tensor as at

X_rv = at.random.normal(0, 1, name="X")
Y_rv = at.random.gamma(0.5, 0.5, name="Y")

p = at.scalar("p")
I_rv = at.random.bernoulli(p, name="I")

M_rv = at.stack([X_rv, Y_rv])[I_rv]

m = M_rv.type()
M_logp = joint_logprob(M_rv, {M_rv: m})
```

Instead of producing a log-probability graph that represents the computation `np.stack([logprob(X, obs), logprob(Y, obs)])[idx]`, which requires complete evaluation of `logprob(X, obs)` and `logprob(Y, obs)` even though `idx` will only select _some_ of those `logprob` values from each mixture component, this implementation will compute something like the following:
```python
res[idx == 0] = logprob(X[idx == 0], obs[idx == 0])
res[idx == 1] = logprob(Y[idx == 1], obs[idx == 1])
```

~Currently, it does all the necessary setup and parsing needed to construct the desired log-probability graphs; however, the actual construction steps aren't necessarily finished.  I still need to check/confirm a few things and&mdash;obviously&mdash;write tests.~

### DIFF
--- a/aeppl/__init__.py
+++ b/aeppl/__init__.py
@@ -4,7 +4,13 @@ __version__ = get_versions()["version"]
 del get_versions
 
 
-from .logprob import logprob  # isort: split
+from aeppl.logprob import logprob  # isort: split
 
-from .joint_logprob import joint_logprob
-from .printing import latex_pprint, pprint
+from aeppl.joint_logprob import joint_logprob
+from aeppl.printing import latex_pprint, pprint
+
+# isort: off
+# Add optimizations to the DBs
+import aeppl.mixture
+
+# isort: on
--- a/aeppl/mixture.py
+++ b/aeppl/mixture.py
@@ -1,0 +1,240 @@
+from typing import List, Optional
+
+import aesara.tensor as at
+import numpy as np
+from aesara.compile.builders import OpFromGraph
+from aesara.graph.basic import Apply
+from aesara.graph.fg import FunctionGraph
+from aesara.graph.opt import local_optimizer, pre_greedy_local_optimizer
+from aesara.ifelse import ifelse
+from aesara.tensor.basic import Join, MakeVector
+from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.random.opt import local_dimshuffle_rv_lift, local_subtensor_rv_lift
+from aesara.tensor.shape import shape_tuple
+from aesara.tensor.type_other import NoneConst
+from aesara.tensor.var import TensorVariable
+
+from aeppl.logprob import _logprob, logprob
+from aeppl.opt import naive_bcast_rv_lift, rv_sinking_db, subtensor_ops
+from aeppl.utils import get_constant_value, indices_from_subtensor
+
+
+def rv_pull_down(x: TensorVariable, dont_touch_vars=None) -> TensorVariable:
+    """Pull a ``RandomVariable`` ``Op`` down through a graph, when possible."""
+    fgraph = FunctionGraph(outputs=dont_touch_vars or [], clone=False)
+
+    return pre_greedy_local_optimizer(
+        fgraph,
+        [
+            local_dimshuffle_rv_lift,
+            local_subtensor_rv_lift,
+            naive_bcast_rv_lift,
+        ],
+        x,
+    )
+
+
+class MixtureRV(OpFromGraph):
+    """A placeholder used to specify a log-likelihood for a mixture sub-graph."""
+
+    default_output = 1
+    # FIXME: This is just to appease `random_make_inplace`
+    inplace = True
+
+    @classmethod
+    def create_node(cls, node, indices, mixture_rvs):
+        out_var = node.default_output()
+
+        inputs = list(indices) + list(mixture_rvs)
+
+        mixture_op = cls(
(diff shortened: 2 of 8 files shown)


aeppl/opt.py
```diff
@@ -178,8 +185,149 @@ def naive_bcast_rv_lift(fgraph, node):
...
+    fgraph.import_node(new_node, import_missing=True, reason="mixture_rv")
+
+    assert new_node.outputs[1].type == out_var.type
+
+    return [new_node.outputs[1]]
+
+
+@_logprob.register(MixtureRV)
+def mixture_logprob(op, value, *inputs, name=None, **kwargs):
+    _, _, _, mixing_rv, *mixture_rvs = inputs
```
ricardoV94, line 306: Is this line needed?

brandonwillard: No, it's not anymore.

aeppl/opt.py
```diff
@@ -178,8 +185,136 @@ def naive_bcast_rv_lift(fgraph, node):
...
+        # Not supported
+        return
+
+    if not isinstance(mixing_rv.owner.op, (BernoulliRV, CategoricalRV)):
+        # Not supported
+        return
+
+    # We index the mixture components with each index element that's in the
+    # component's group
+    value_var = rv_map_feature.rv_values.pop(out_var)
```
ricardoV94, line 267: What if the user did not specify a value var to the mixture (because they don't want to)? This would raise a KeyError, no?

brandonwillard: Yes; I'll update that.

brandonwillard:
The multivariate case is blocked by https://github.com/aesara-devs/aesara/pull/507 and https://github.com/aesara-devs/aesara/pull/509.

The latter issue is only a blocker if we want to prevent the evaluation of "unselected" mixture components.  We can always produce an array of stacked `logprob` values evaluated conditional on mixture component and _then_ index that array with the mixing distribution; however, this is a prohibitively slow approach for large enough mixture models (e.g. almost anything involving time-varying mixture components used with non-trivially-sized datasets)

brandonwillard:
The two tests in this PR pass locally under the changes in https://github.com/aesara-devs/aesara/pull/516.

aeppl/__init__.py
```diff
@@ -4,7 +4,13 @@
...
-from .logprob import logprob  # isort: split
+from aeppl.logprob import logprob  # isort: split
 
-from .joint_logprob import joint_logprob
-from .printing import latex_pprint, pprint
+from aeppl.joint_logprob import joint_logprob
+from aeppl.printing import latex_pprint, pprint
+
+# isort: off
+# Add optimizations to the DBs
```
ricardoV94, line 13: Is this comment needed?

aeppl/mixture.py
```diff
@@ -0,0 +1,237 @@
...
+
+        if join_axis != 0:
+            # TODO: Support other join axes
+            return
+
+    if not all(
+        rv.owner and isinstance(rv.owner.op, RandomVariable) for rv in mixture_rvs
+    ):
+        # Currently, all mixture components must be `RandomVariable` outputs
+        # TODO: Allow constants and make them Dirac-deltas
```
ricardoV94, line 120: We could have a rewrite phase before this one where we convert constants to Dirac deltas (not needed for this PR)

aeppl/mixture.py
```diff
@@ -0,0 +1,237 @@
...
+        getattr(subtensor_node.op, "idx_list", None), indices
+    )
+    comp_rvs = inputs[num_indices:]
+
+    if value.ndim > 0:
+        # TODO: Make the join axis to the left-most dimension (or transpose the
+        # problem)
+        join_axis = 0  # op.join_axis
+
+        logp_val = at.full(tuple(at.shape(value)), -np.inf, dtype=value.dtype)
```
ricardoV94, line 193: Wouldn't be safer to fill it with `nan`s?

ricardoV94:
Is this working with categorical distributions as well? If so, we could add a test

aeppl/__init__.py
brandonwillard: Why wouldn't it be?

aeppl/mixture.py
brandonwillard: Safer in what way?  It probably shouldn't matter what that initial value is, since the indexing is supposed to cover all values.

brandonwillard:
> Is this working with categorical distributions as well? If so, we could add a test

Yes, definitely.

aeppl/__init__.py
ricardoV94: It was unclear to me if it meant anything

brandonwillard: Oh, it's just a warning that the import is needed in order to register the mixture optimizations, and it's not there to expose the mixture module at the package level, for example.

PR #22, opened 2021-06-22
Censored RVs

ricardoV94:
This PR implements logprob for censored (clipped) RVs. 

```python
x_rv = at.random.normal(0, 1)
cens_x_rv = at.clip(x_rv, -1, 1)
cens_x = cens_x_rv.type()
logp = ppl.joint_logprob({cens_x_rv: cens_x})
```

I placed the new methods and tests inside `truncation.py`, expecting this file wil also contain the methods for truncated RVs in the future.

***

Some things are still not working well / missing:

+ [x] Issues with broadcasting / sizes (see xfailed tests)
+ [x] Propagate names in logp graph
+ [x] ~Canonicalize set_subtensors to clip `x[x>ub] = ub -> clip(x, x, ub)`~ Will do in another PR
+ [x] Add tests for `logcdf` methods
+ [x] ~Compute test values for new nodes~ Seems to not be necessary, tests pass with `compute_test_value="raise"`
+ [x] ~Explore if CensoredRVs should be created even when they don't have a direct value variable (e.g, so that they can work as input to other derivedRVs)~ Postponed

### DIFF
--- a/aeppl/truncation.py
+++ b/aeppl/truncation.py
@@ -1,0 +1,126 @@
+import warnings
+from typing import List, Optional
+
+import aesara.tensor as at
+import numpy as np
+from aesara.assert_op import Assert
+from aesara.graph.basic import Node
+from aesara.graph.fg import FunctionGraph
+from aesara.graph.opt import local_optimizer
+from aesara.scalar.basic import Clip
+from aesara.scalar.basic import clip as scalar_clip
+from aesara.tensor.elemwise import Elemwise
+from aesara.tensor.var import TensorConstant
+
+from aeppl.abstract import MeasurableVariable, assign_custom_measurable_outputs
+from aeppl.logprob import _logcdf, _logprob
+from aeppl.opt import rv_sinking_db
+
+
+class CensoredRV(Elemwise):
+    """A placeholder used to specify a log-likelihood for a censored RV sub-graph."""
+
+
+MeasurableVariable.register(CensoredRV)
+
+
+@local_optimizer(tracks=[Elemwise])
+def find_censored_rvs(fgraph: FunctionGraph, node: Node) -> Optional[List[CensoredRV]]:
+    # TODO: Canonicalize x[x>ub] = ub -> clip(x, x, ub)
+
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    if isinstance(node.op, CensoredRV):
+        return None  # pragma: no cover
+
+    if not (isinstance(node.op, Elemwise) and isinstance(node.op.scalar_op, Clip)):
+        return None
+
+    clipped_var = node.outputs[0]
+    if clipped_var not in rv_map_feature.rv_values:
+        return None
+
+    base_var, lower_bound, upper_bound = node.inputs
+
+    if not (base_var.owner and isinstance(base_var.owner.op, MeasurableVariable)):
+        return None
+
+    if base_var in rv_map_feature.rv_values:
+        warnings.warn(
+            f"Value variables were assigned to both the input ({base_var}) and "
+            f"output ({clipped_var}) of a censored random variable."
+        )
+        return None
+
+    # Replace bounds by `+-inf` if `y = clip(x, x, ?)` or `y=clip(x, ?, x)`
+    # This is used in `censor_logprob` to generate a more succint logprob graph
+    # for one-sided censored random variables
+    lower_bound = lower_bound if (lower_bound is not base_var) else at.constant(-np.inf)
+    upper_bound = upper_bound if (upper_bound is not base_var) else at.constant(np.inf)
+
+    censored_op = CensoredRV(scalar_clip)
+    # Make base_var unmeasurable
+    unmeasurable_base_var = assign_custom_measurable_outputs(base_var.owner)
+    censored_rv_node = censored_op.make_node(
+        unmeasurable_base_var, lower_bound, upper_bound
+    )
+    censored_rv = censored_rv_node.outputs[0]
+
+    censored_rv.name = clipped_var.name
+
+    return [censored_rv]
+
+
+rv_sinking_db.register("find_censored_rvs", find_censored_rvs, -5, "basic")
+
+
+@_logprob.register(CensoredRV)
+def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
+    (value,) = values
+
+    base_rv_op = base_rv.owner.op
+    base_rv_inputs = base_rv.owner.inputs
+
+    logprob = _logprob(base_rv_op, (value,), *base_rv_inputs, **kwargs)
+    logcdf = _logcdf(base_rv_op, value, *base_rv_inputs, **kwargs)
+
+    if base_rv_op.name:
+        logprob.name = f"{base_rv_op}_logprob"
+        logcdf.name = f"{base_rv_op}_logcdf"
+
+    is_lower_bounded, is_upper_bounded = False, False
+    if not (
+        isinstance(upper_bound, TensorConstant) and np.all(np.isinf(upper_bound.value))
+    ):
+        is_upper_bounded = True
+
+        logccdf = at.log(1 - at.exp(logcdf))
+        # For right censored discrete RVs, we need to add an extra term
+        # corresponding to the pmf at the upper bound
+        if base_rv_op.dtype == "int64":
+            logccdf = at.logaddexp(logccdf, logprob)
+
+        logprob = at.switch(
+            at.eq(value, upper_bound),
+            logccdf,
+            at.switch(at.gt(value, upper_bound), -np.inf, logprob),
+        )
+    if not (
+        isinstance(lower_bound, TensorConstant)
+        and np.all(np.isneginf(lower_bound.value))
+    ):
+        is_lower_bounded = True
+        logprob = at.switch(
+            at.eq(value, lower_bound),
+            logcdf,
+            at.switch(at.lt(value, lower_bound), -np.inf, logprob),
+        )
+
+    if is_lower_bounded and is_upper_bounded:
+        logprob = Assert("lower_bound <= upper_bound")(
+            logprob, at.all(at.le(lower_bound, upper_bound))
+        )
+
+    return logprob
--- a/tests/test_truncation.py
+++ b/tests/test_truncation.py
@@ -1,0 +1,192 @@
+import aesara
+import aesara.tensor as at
+import numpy as np
+import pytest
+import scipy as sp
+import scipy.stats as st
+
+from aeppl import factorized_joint_logprob, joint_logprob
+from aeppl.transforms import LogTransform, TransformValuesOpt
+from tests.utils import assert_no_rvs
+
+
+@aesara.config.change_flags(compute_test_value="raise")
+def test_continuous_rv_censoring():
+    x_rv = at.random.normal(0.5, 1)
+    cens_x_rv = at.clip(x_rv, -2, 2)
+
+    cens_x_vv = cens_x_rv.clone()
+    cens_x_vv.tag.test_value = 0
+
+    logp = joint_logprob({cens_x_rv: cens_x_vv})
+    assert_no_rvs(logp)
+
+    logp_fn = aesara.function([cens_x_vv], logp)
+    ref_scipy = st.norm(0.5, 1)
+
+    assert logp_fn(-3) == -np.inf
+    assert logp_fn(3) == -np.inf
+
+    assert np.isclose(logp_fn(-2), ref_scipy.logcdf(-2))
+    assert np.isclose(logp_fn(2), ref_scipy.logsf(2))
+    assert np.isclose(logp_fn(0), ref_scipy.logpdf(0))
+
+
+def test_discrete_rv_censoring():
+    x_rv = at.random.poisson(2)
+    cens_x_rv = at.clip(x_rv, 1, 4)
+
+    cens_x_vv = cens_x_rv.clone()
+
+    logp = joint_logprob({cens_x_rv: cens_x_vv})
+    assert_no_rvs(logp)
+
+    logp_fn = aesara.function([cens_x_vv], logp)
+    ref_scipy = st.poisson(2)
+
+    assert logp_fn(0) == -np.inf
+    assert logp_fn(5) == -np.inf
+
+    assert np.isclose(logp_fn(1), ref_scipy.logcdf(1))
+    assert np.isclose(logp_fn(4), np.logaddexp(ref_scipy.logsf(4), ref_scipy.logpmf(4)))
+    assert np.isclose(logp_fn(2), ref_scipy.logpmf(2))
+
+
+def test_one_sided_censoring():
+    x_rv = at.random.normal(0, 1)
+    lb_cens_x_rv = at.clip(x_rv, -1, x_rv)
+    ub_cens_x_rv = at.clip(x_rv, x_rv, 1)
+
+    lb_cens_x_vv = lb_cens_x_rv.clone()
+    ub_cens_x_vv = ub_cens_x_rv.clone()
+
+    lb_logp = joint_logprob({lb_cens_x_rv: lb_cens_x_vv})
+    ub_logp = joint_logprob({ub_cens_x_rv: ub_cens_x_vv})
+    assert_no_rvs(lb_logp)
+    assert_no_rvs(ub_logp)
+
+    logp_fn = aesara.function([lb_cens_x_vv, ub_cens_x_vv], [lb_logp, ub_logp])
+    ref_scipy = st.norm(0, 1)
+
+    assert np.all(np.array(logp_fn(-2, 2)) == -np.inf)
+    assert np.all(np.array(logp_fn(2, -2)) != -np.inf)
+    np.testing.assert_almost_equal(logp_fn(-1, 1), ref_scipy.logcdf(-1))
+    np.testing.assert_almost_equal(logp_fn(1, -1), ref_scipy.logpdf(-1))
+
+
+def test_useless_censoring():
+    x_rv = at.random.normal(0.5, 1, size=3)
+    cens_x_rv = at.clip(x_rv, x_rv, x_rv)
+
+    cens_x_vv = cens_x_rv.clone()
+
+    logp = joint_logprob({cens_x_rv: cens_x_vv}, sum=False)
+    assert_no_rvs(logp)
+
+    logp_fn = aesara.function([cens_x_vv], logp)
+    ref_scipy = st.norm(0.5, 1)
+
+    np.testing.assert_allclose(logp_fn([-2, 0, 2]), ref_scipy.logpdf([-2, 0, 2]))
+
+
+def test_random_censoring():
+    lb_rv = at.random.normal(0, 1, size=2)
+    x_rv = at.random.normal(0, 2)
+    cens_x_rv = at.clip(x_rv, lb_rv, [1, 1])
+
+    lb_vv = lb_rv.clone()
+    cens_x_vv = cens_x_rv.clone()
+    logp = joint_logprob({cens_x_rv: cens_x_vv, lb_rv: lb_vv}, sum=False)
+    assert_no_rvs(logp)
+
+    logp_fn = aesara.function([lb_vv, cens_x_vv], logp)
+    res = logp_fn([0, -1], [-1, -1])
+    assert res[0] == -np.inf
+    assert res[1] != -np.inf
+
+
+def test_broadcasted_censoring_constant():
+    lb_rv = at.random.uniform(0, 1)
+    x_rv = at.random.normal(0, 2)
+    cens_x_rv = at.clip(x_rv, lb_rv, [1, 1])
+
+    lb_vv = lb_rv.clone()
+    cens_x_vv = cens_x_rv.clone()
+
+    logp = joint_logprob({cens_x_rv: cens_x_vv, lb_rv: lb_vv})
+    assert_no_rvs(logp)
+
+
+def test_broadcasted_censoring_random():
+    lb_rv = at.random.normal(0, 1)
+    x_rv = at.random.normal(0, 2, size=2)
+    cens_x_rv = at.clip(x_rv, lb_rv, 1)
+
+    lb_vv = lb_rv.clone()
+    cens_x_vv = cens_x_rv.clone()
+
+    logp = joint_logprob({cens_x_rv: cens_x_vv, lb_rv: lb_vv})
+    assert_no_rvs(logp)
+
+
+def test_fail_base_and_censored_have_values():
+    """Test failure when both base_rv and clipped_rv are given value vars"""
+    x_rv = at.random.normal(0, 1)
+    cens_x_rv = at.clip(x_rv, x_rv, 1)
+    cens_x_rv.name = "cens_x"
+
+    x_vv = x_rv.clone()
+    cens_x_vv = cens_x_rv.clone()
+    with pytest.warns(UserWarning):
+        logp_terms = factorized_joint_logprob({cens_x_rv: cens_x_vv, x_rv: x_vv})
+    assert cens_x_vv not in logp_terms
+
+
+def test_fail_multiple_censored_single_base():
+    """Test failure when multiple clipped_rvs share a single base_rv"""
+    base_rv = at.random.normal(0, 1)
+    cens_rv1 = at.clip(base_rv, -1, 1)
+    cens_rv1.name = "cens1"
+    cens_rv2 = at.clip(base_rv, -1, 1)
+    cens_rv2.name = "cens2"
+
+    cens_vv1 = cens_rv1.clone()
+    cens_vv2 = cens_rv2.clone()
+    logp_terms = factorized_joint_logprob({cens_rv1: cens_vv1, cens_rv2: cens_vv2})
+    assert cens_rv2 not in logp_terms
+
+
+def test_deterministic_clipping():
+    x_rv = at.random.normal(0, 1)
+    clip = at.clip(x_rv, 0, 0)
+    y_rv = at.random.normal(clip, 1)
+
+    x_vv = x_rv.clone()
+    y_vv = y_rv.clone()
+    logp = joint_logprob({x_rv: x_vv, y_rv: y_vv})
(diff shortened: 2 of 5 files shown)


aeppl/truncation.py
```diff
@@ -0,0 +1,95 @@
...
+    if not isinstance(node.op, Elemwise) or not isinstance(node.op.scalar_op, Clip):
+        return
+
+    base_var, lower_bound, upper_bound = node.inputs
+    base_op = base_var.owner.op
+
+    if not isinstance(base_op, RandomVariable):
+        return
+
+    # clipped_var = node.ouputs[0]
```
ricardoV94, line 77: For some weird reason calling `node.outputs` fails, saying `Apply` does not have outputs attribute.

aeppl/truncation.py
```diff
@@ -0,0 +1,95 @@
...
+    # clipped_value = rv_map_feature.rv_values.pop(clipped_var)
+
+    censored_rv = CensoredRV(
+        "censored",
+        base_op.ndim_supp,
+        list(base_op.ndims_params) + [0, 0],
```
ricardoV94, line 83:
```suggestion
        list(base_op.ndims_params) + [base_op.ndim_supp]*2,
```

aeppl/truncation.py
```diff
@@ -0,0 +1,95 @@
...
+
+@local_optimizer(tracks=[Elemwise])
+def censor_rvs(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]:
+
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return
+
+    # TODO: Allow for one sided-truncation with set_subtensor graph (x[x>ub] = ub)
```
ricardoV94, line 67: Maybe the one sided censoring can be "canonicalized" to a `x = clip(x, x, ub)` and `x = clip(x, lb, x)`?

ricardoV94:
```suggestion
    # TODO: Allow for one sided-censoring with set_subtensor graph (x[x>ub] = ub)
```

aeppl/truncation.py
ricardoV94: Apparently for `Elemwise` nodes it should be `out` and not `outputs`? Also, is there a way to target the `Clip` op directly and not the `Elemwise`?

aeppl/truncation.py
```diff
@@ -0,0 +1,121 @@
...
+        list(base_op.ndims_params) + [0, 0],
+        base_op.dtype,
+        inplace=False,
+        base_op=base_op,
+    )
+
+    censored_node = censored_rv.make_node(
+        *base_var.owner.inputs,
+        lower_bound if is_lower_bounded else -np.inf,
+        upper_bound if is_upper_bounded else np.inf,
```
ricardoV94, line 115: These lines look problematic: `-np.inf` should have the same shape as `[lower|upper]_bound`

tests/test_truncation.py
```diff
@@ -0,0 +1,124 @@
...
+
+    lb_logp = joint_logprob(lb_cens_x_rv, {lb_cens_x_rv: lb_cens_x})
+    ub_logp = joint_logprob(ub_cens_x_rv, {ub_cens_x_rv: ub_cens_x})
+    assert_no_rvs(lb_logp)
+    assert_no_rvs(ub_logp)
+
+    logp_fn = aesara.function([lb_cens_x, ub_cens_x], [lb_logp, ub_logp])
+
+    assert np.all(np.array(logp_fn(-2, 2)) == -np.inf)
+    assert np.all(np.array(logp_fn(2, -2)) != -np.inf)
```
ricardoV94, line 68: Todo assert actual logp in valid domain (and at edge)

aeppl/truncation.py
brandonwillard: You've sorted this one out, no?

brandonwillard: > Apparently for `Elemwise` nodes it should be `out` and not `outputs`?

The `outputs` attribute should always be available on an `Apply` node.  If not, there's a serious bug somewhere.

brandonwillard: > is there a way to target the `Clip` Op directly instead of the `Elemwise`?

If only one instance of the `Elemwise` + `Clip` combination is used (i.e. it's a singleton), then you can target that instance directly; otherwise, no, there's currently no easy (e.g. a single `isinstance`) type-based way to identify this combination.

If we implement `Op` singletons at the metaclass level, this situation could be simplified a lot.  Both alternatively and in addition to that, we could consider overriding `__instancecheck__` so that `isinstance(node.op, at.clip)` will check `issubclass(type(node.op), Elemwise) and isinstance(node.op.scalar_op, Clip)` (or perhaps use strict equality).

brandonwillard:
To make broadcasting work, you should be able to use something like `at.broadcast_arrays(value, lower_bound, upper_bound)` in the log-probability calculation.

brandonwillard:
> * Canonicalize set_subtensors to clip `x[x>ub] = ub -> clip(x, x, ub)`

That's a very interesting idea!  Can it be combined with other indices (e.g. `x[idx_1, ..., x > ub, ..., idx_N] = ...`)?

Regardless, don't make this PR conditional on such extensions.  Let's get `at.clip` working and merged first.

brandonwillard:
Tip: if you assign names to your test `Variable`s, it can be much easier to read the `aesara.dprint` output when debugging.  Same with `Variable`s created in the functions you're working on (e.g. in `censor_logprob` and/or `censor_rvs`).

Also, don't forget about test values!  They will cause errors to arise during graph construction (i.e. where the symbolic objects themselves are defined).  That combined with `aesara.config.print_test_value = True` can make things a lot easier.

aeppl/truncation.py
ricardoV94: > > Apparently for `Elemwise` nodes it should be `out` and not `outputs`?
> 
> The `outputs` attribute should always be available on an `Apply` node. If not, there's a serious bug somewhere.

I might have confused and typed `output` singular without realizing it (but type the correct `outputs` during interactive debugging, which confused me). Things seem fine!

ricardoV94: > If we implement `Op` singletons at the metaclass level, this situation could be simplified a lot. Both alternatively and in addition to that, we could consider overriding `__subclasshook__`/`__subclasscheck__` so that `issubclass(type(node.op), at.clip)` will check `issubclass(type(node.op), Elemwise) and isinstance(type(node.op).scalar_op, Clip)` (or perhaps use strict equality). We would need to make `scalar_op` a class-level attribute, though.

That sounds really useful. In most of the rewrites / opts, I find myself having to do these instance checks manually

ricardoV94:
> To make broadcasting work, you should be able to use something like `at.broadcast_arrays(value, lower_bound, upper_bound)` in the log-probability calculation.

Still couldn't figure out my `test_broadcasted_censoring` failing test. These are the fgraphs printouts before and after the optimization phase in `joint_logprob`:

```Elemwise{clip,no_inplace} [id A] 'cens_x_rv'   4
 |InplaceDimShuffle{x} [id B] ''   3
 | |normal_rv.1 [id C] 'x_rv'   2
 |   |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F64081A3440>) [id D]
 |   |TensorConstant{[]} [id E]
 |   |TensorConstant{11} [id F]
 |   |TensorConstant{0} [id G]
 |   |TensorConstant{2} [id H]
 |InplaceDimShuffle{x} [id I] ''   1
 | |uniform_rv.1 [id J] 'lb_rv'   0
 |   |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F642C9B3A40>) [id K]
 |   |TensorConstant{[]} [id L]
 |   |TensorConstant{11} [id M]
 |   |TensorConstant{0} [id N]
 |   |TensorConstant{1} [id O]
 |TensorConstant{(2,) of 1} [id P]

After:
censored_rv.1 [id A] 'cens_x_rv'   5
 |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F64081A3440>) [id B]
 |TensorConstant{[]} [id C]
 |TensorConstant{11} [id D]
 |InplaceDimShuffle{x} [id E] ''   4
 | |TensorConstant{0} [id F]
 |InplaceDimShuffle{x} [id G] ''   3
 | |TensorConstant{2} [id H]
 |uniform_rv.1 [id I] ''   2
 | |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F642C9B3A40>) [id J]
 | |TensorConstant{[]} [id K]
 | |TensorConstant{11} [id L]
 | |InplaceDimShuffle{x} [id M] ''   1
 | | |TensorConstant{0} [id N]
 | |InplaceDimShuffle{x} [id O] ''   0
 |   |TensorConstant{1} [id P]
 |TensorConstant{(2,) of 1} [id Q]
```

I seem to be loosing the reference to the `lb_rv` before I even get to my `censor_rvs` opt. This is caused by the `local_dimshuffle_rv_lift`. During interactive debugging:

```python
lower_bound.owner.tag
# scratchpad{'imported_by': ['local_dimshuffle_rv_lift']}
```

Also that opt seems to not propagate the variable name

```python
lower_bound.name
# None
```

ricardoV94:
> Also, don't forget about test values! They will cause errors to arise during graph construction (i.e. where the symbolic objects themselves are defined). That combined with `aesara.config.print_test_value = True` can make things a lot easier.

If I set tthe flag to "warn", even a simple unform logp raises a lot of "Cannot compute test value..." for every node in the logp. Is this something we need to address?

```python
@aesara.config.change_flags(compute_test_value='warn')
def test_compute_test_value():
    x_rv = at.random.uniform(-1, 1)
    x = x_rv.type()
    logp = joint_logprob(x_rv, {x_rv: x})
```
```python
=============================== warnings summary ===============================
tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/venv/lib/python3.8/site-packages/aesara/graph/op.py:272: UserWarning: Warning, Cannot compute test value: input 0 (<TensorType(float64, scalar)>) of Op Elemwise{ge,no_inplace}(<TensorType(float64, scalar)>, TensorConstant{-1}) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/venv/lib/python3.8/site-packages/aesara/graph/op.py:272: UserWarning: Warning, Cannot compute test value: input 0 (<TensorType(float64, scalar)>) of Op Elemwise{le,no_inplace}(<TensorType(float64, scalar)>, TensorConstant{1}) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/venv/lib/python3.8/site-packages/aesara/graph/op.py:272: UserWarning: Warning, Cannot compute test value: input 0 (Elemwise{ge,no_inplace}.0) of Op Elemwise{and_,no_inplace}(Elemwise{ge,no_inplace}.0, Elemwise{le,no_inplace}.0) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/venv/lib/python3.8/site-packages/aesara/graph/op.py:272: UserWarning: Warning, Cannot compute test value: input 0 (<TensorType(float64, scalar)>) of Op Elemwise{second,no_inplace}(<TensorType(float64, scalar)>, Elemwise{neg,no_inplace}.0) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/venv/lib/python3.8/site-packages/aesara/graph/op.py:272: UserWarning: Warning, Cannot compute test value: input 0 (Elemwise{and_,no_inplace}.0) of Op Elemwise{switch,no_inplace}(Elemwise{and_,no_inplace}.0, Elemwise{second,no_inplace}.0, TensorConstant{-inf}) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/aeppl/joint_logprob.py:170: UserWarning: Warning, Cannot compute test value: input 0 (<TensorType(float64, scalar)>) of Op Elemwise{second,no_inplace}(<TensorType(float64, scalar)>, Elemwise{neg,no_inplace}.0) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/aeppl/joint_logprob.py:170: UserWarning: Warning, Cannot compute test value: input 0 (<TensorType(float64, scalar)>) of Op Elemwise{le,no_inplace}(<TensorType(float64, scalar)>, TensorConstant{1}) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/aeppl/joint_logprob.py:170: UserWarning: Warning, Cannot compute test value: input 0 (<TensorType(float64, scalar)>) of Op Elemwise{ge,no_inplace}(<TensorType(float64, scalar)>, TensorConstant{-1}) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/aeppl/joint_logprob.py:170: UserWarning: Warning, Cannot compute test value: input 0 (Elemwise{ge,no_inplace}.0) of Op Elemwise{and_,no_inplace}(Elemwise{ge,no_inplace}.0, Elemwise{le,no_inplace}.0) missing default value
    compute_test_value(node)

tests/test_truncation.py::test_compute_test_value
  /home/ricardo/Documents/Projects/aeppl/aeppl/joint_logprob.py:170: UserWarning: Warning, Cannot compute test value: input 0 (Elemwise{and_,no_inplace}.0) of Op Elemwise{switch,no_inplace}(Elemwise{and_,no_inplace}.0, Elemwise{second,no_inplace}.0, TensorConstant{-inf}) missing default value
    compute_test_value(node)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================== 1 passed, 10 warnings in 0.97s ========================

Process finished with exit code 0
PASSED                       [100%]
```

brandonwillard:
> Is this something we need to address?

It looks like you need to set s test value for `x`, or use `x_rv.clone()`.

ricardoV94:
I added tests for the logcdf methods and checked whether the new opts work with `compute_test_value="raise`. 

The only thing missing are the broadcasting / `local_dimshuffle_rv_lift` related issues I described in https://github.com/aesara-devs/aeppl/pull/22#issuecomment-870377413

aeppl/truncation.py
```diff
@@ -0,0 +1,126 @@
...
+    censored_node = censored_rv.make_node(
+        *base_var.owner.inputs,
+        lower_bound,
+        upper_bound,
+    )
+
+    censored_node_out = censored_node.outputs[1]
+
+    if not censored_node_out.type == clipped_var.type:
+        return
```
brandonwillard, line 119: This would make more sense as an `assert`.  Even so, this shouldn't be a necessary check, because the logic that determines the output type is actually in `RandomVariable` and is specified directly in the above `CensoredRV` constructor call.

NB: You might be able to simplify the use of `CensoredRV` by extracting the `dtype`&mdash;and other poperties from `base_op`&mdash;in `CensoredRV.__init__`.

aeppl/truncation.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(self, *args, base_op, **kwargs):
+        self.base_op = base_op
+        super().__init__(*args, **kwargs)
+
+
+@_logprob.register(CensoredRV)
+def censor_logprob(op, value, *inputs, name=None, **kwargs):
+
+    *rv_params, lower_bound, upper_bound = inputs
+    logprob = _logprob(op.base_op, value, *rv_params, name=name, **kwargs)
```
brandonwillard, line 31: You could alternatively leave the original `RandomVariable` in the graph, but make in an input of a `CensoredRV`, and this log-probability term would be taken care of automatically, then you can use this dispatch function to add the extra log-probability terms introduced by censoring.

brandonwillard:
> That `uniform_rv.1` in the "After graph" is no longer related to the original one.
> This is caused by the `local_dimshuffle_rv_lift`. During interactive debugging:
> 
> ```python
> lower_bound
> # uniform_rv.out
> 
> lower_bound in rv_map_feature.rv_values
> # False
> 
> lower_bound.owner.tag
> # scratchpad{'imported_by': ['local_dimshuffle_rv_lift']}
> ```
> 
> Also that opt seems to not propagate the variable name
> 
> ```python
> lower_bound.name
> # None
> ```

Ah, yeah, that would definitely cause such a problem; however, `PreserveRVMappings` is supposed to take care of that, so the question I would ask is "Why isn't `PreserveRVMappings.on_change_input` catching this case?".

aeppl/truncation.py
ricardoV94: In that case I will have to direct the `value` var to the original variable in the rewrite phase (while keeping it pointing to the `censor_logprob` of course). I can give it a try, but I am not sure it will result in simpler code.

aeppl/truncation.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+
+    return logprob
+
+
+@local_optimizer(tracks=[Elemwise])
+def censor_rvs(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]:
+    # TODO: Canonicalize x[x>ub] = ub -> clip(x, x, ub)
+
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
```
ricardoV94, line 83: Can we just let this kind of check fail instead of using the `getattr`?

aeppl/truncation.py
brandonwillard: After #19 is merged, you could also use `.tag.ignore_logprob`.  With that, we can write a general `RandomVariable` wrapper/view `Op` that can be used as the base class for something like `CensoredRV`.

aeppl/truncation.py
```diff
@@ -0,0 +1,129 @@
...
+    is_lower_bounded, is_upper_bounded = False, False
+    if not (
+        isinstance(upper_bound, TensorConstant) and np.all(upper_bound.value == np.inf)
+    ):
+        is_upper_bounded = True
+
+        logccdf = at.log(1 - at.exp(logcdf))
+        # For right censored discrete RVs, we need to add an extra term
+        # corresponding to the pmf at the upper bound
+        if op.base_op.dtype == "int64":
```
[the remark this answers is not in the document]
brandonwillard: We can create a metatype that does these kinds of checks.

aeppl/truncation.py
brandonwillard: No, we need to return `None` to signify that the optimization is skipped.  I think the optimization framework will allow uncaught exceptions, but they'll show up as warnings, and we don't want that.  

Plus, I believe there are still a few places where exceptions are used for control flow during optimization, and we still need to fix that.

tests/test_truncation.py
```diff
@@ -0,0 +1,177 @@
...
+    assert_no_rvs(logp)
+
+    logp_fn = aesara.function([lb, cens_x], logp)
+    res = logp_fn([0, -1], [-1, -1])
+    assert res[0] == -np.inf
+    assert res[1] != -np.inf
+
+
+@pytest.mark.xfail(reason="Broadcasting not properly handled yet")
+def test_broadcasted_censoring_constant():
```
ricardoV94, line 106: I am okay with not supporting this case where the base RV does not have the right size, but the one in `test_broadcasted_censoring_random` below should work fine.

ricardoV94:
Thinking a bit more about the generating models in the broadcasted test examples I wrote, I think it only makes sense to create a logprob graph when the base random variable has the same shape as the clipped output.

```python
# Scalar base, vector random bound - SUCCEEDS
lb_rv = at.random.normal(0, 1, size=2)
x_rv = at.random.normal(0, 2)
cens_x_rv = at.clip(x_rv, lb_rv, [1, 1])

# Scalar base, vector bound - FAILS
lb_rv = at.random.uniform(0, 1, name="lb_rv")
x_rv = at.random.normal(0, 2, name="x_rv")
cens_x_rv = at.clip(x_rv, lb_rv, [1, 1])

# Vector base RV, vector bound - FAILS
lb_rv = at.random.normal(0, 1, name="lb_rv")
x_rv = at.random.normal(0, 2, size=2, name="x_rv")
cens_x_rv = at.clip(x_rv, lb_rv, [1, 1])
```
The logprob extraction for the first two graphs should fail because the two (generated) values in `cens_x_rv` are not independent (they are just broadcasted). In contrast, the logprob extraction for last graph should succeed, since each observation corresponds to a new independent factor. 

Is there a way to distinguish between these during the rewrite phase?

brandonwillard:
> Is there a way to distinguish between these during the rewrite phase?

Perhaps we could check the `RandomVariable.ndim_supp` and broadcasted shape/number of dimensions of the censored variable and compare that with the other clipped arguments.  We know that `Clip` will simply broadcast all its arguments, so, when the censored variable is a set of independent variates&mdash;e.g. when `size` is non-`None` or multiple independent distribution parameters are provided&mdash;that matches the broadcast dimensions of `Clip`'s other arguments, we can probably say that the censoring itself is broadcasted (and not "correlated").

aeppl/joint_logprob.py
```diff
@@ -213,6 +213,20 @@ def factorized_joint_logprob(
...
                 compute_test_value(node)
 
+    missing_value_terms = set(original_values.values()) - set(logprob_vars.keys())
+    if missing_value_terms:
+        # Filter out missing terms of variables with ignore_logprob
+        value_rvs = {v: k for k, v in updated_rv_values.items()}
+        for missing in tuple(missing_value_terms):
+            rv_of_missing = value_rvs.get(missing, None)
+            if rv_of_missing and getattr(rv_of_missing.tag, "ignore_logprob", False):
+                missing_value_terms.remove(missing)
```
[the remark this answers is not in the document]
brandonwillard: > When I said something could be removed later, I meant the specific ignore_logprob flag logic, not the raising a RuntimeError if a variable is missing.
> 
> While developing the censored variables it would often fail silently and just return a graph with aesara clips unchanged and/or less terms than requested. This seems like a good way to catch such failures.

From a simple development and design perspective, it sounds like you're addressing a testing-specific issue within a feature implementation, and that's generally not good.

Otherwise, if something is failing silently the first question is "What's failing?".  Is it the `factorized_joint_logprob` loop?  If not, the failure should be addressed closer to where its primary logic/code resides, and that doesn't appear to be here.

> Is there a reason why you don't want this type of check at the end of `factorized_joint_logprob`?
> 
> It's trivial to manually do the same check in those new tests I added, but this explicit check might be valuable enough to have as a default.

The reason why I don't want these kinds of unrelated changes is that their inclusion makes a PR contingent on additional review work and discussions.

It takes extra time and effort to go through logic like this and determine its relevance, risk, etc.  These are things that need to be done within issues and/or at the outset of a PR (e.g. the premise/description of a PR) in order to avoid delaying the inclusion of any agreed upon changes.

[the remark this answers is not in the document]
ricardoV94: > From a simple development and design perspective, it sounds like you're addressing a testing-specific issue within a feature implementation, and that's generally not good.

What happened was that the test revealed that such a call to factorized_logprob  would return with a missing logp term and zero complaints so I decided to add an explicit check there. 

It was not for the sake of the test as that had already been solved and could be tested explicitly inside the test itself.

It was meant for further development when we introduce rewrites for Ops that are otherwise valid in logp graphs. It's also a conceptual obvious check for me: a user requested a dictionary of rv_values and we make sure we are returning a dictionary with a item for each original pair.

I don't mind splitting this into another PR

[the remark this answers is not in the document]
ricardoV94: I removed the commit but I am not very happy with the fact that this does not fail naturally:

```python
def test_fail_base_and_censored_have_values():
    """Test failure when both base_rv and clipped_rv are given value vars"""
    x_rv = at.random.normal(0, 1)
    cens_x_rv = at.clip(x_rv, x_rv, 1)
    cens_x_rv.name = "cens_x"

    x_vv = x_rv.clone()
    cens_x_vv = cens_x_rv.clone()
    logp_terms = factorized_joint_logprob({cens_x_rv: cens_x_vv, x_rv: x_vv})
    assert cens_x_vv not in logp_terms
```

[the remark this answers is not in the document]
brandonwillard: Again, why should the error be raised in `factorized_joint_logprob` specifically?  

The two terms involved are very specific to the censored variable logic, and it looks like the error could've been initiated in `find_censored_rvs`&mdash;i.e. where all the relevant terms are identified and used.  This approach could also short-circuit all the unnecessary down-stream logic, no?

brandonwillard (commented):
You can use the same tests as before, but change the `pytest.raises` to look for the warnings instead.

aeppl/joint_logprob.py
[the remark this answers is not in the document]
ricardoV94: I don't think it should be the rewrite responsibility to raise a failure. It may be there is another rewrite (e.g., added by users of the library) that can handle the conversion.

The bigger problem is that nothing happens if you ask for a graph that we don't know how to handle. That is not specific to censoredRVs, we just haven't tested it. For instance, this snippet does not complain at all:

```python
import aesara.tensor as at
import aeppl

x_rv = at.random.normal(name='x')
y_rv = at.cos(x_rv)

x_vv = x_rv.clone()
y_vv = y_rv.clone()

logprob_dict = aeppl.factorized_joint_logprob({x_rv: x_vv, y_rv: y_vv})
logprob_dict
{x: x_logprob}
```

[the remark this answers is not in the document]
brandonwillard: > The bigger problem is that nothing happens if you ask for a graph that we don't know how to handle. That is not specific to censoredRVs, we just haven't tested it. For instance, this snippet does not complain at all:

If our rewrites don't know how to handle something, that's not necessarily a problem.  As a matter of fact, we _expect_ that they won't know how to handle more things than they do.

The assumption underlying your statements and example seems to be that you know what should be done **relative to specific rewrites**, and this is what makes it reasonable to handle rewrite-relevant errors in the rewrite logic.  

In other words, if you know an error/warning should be raised because a value variable specification is redundant, you only really know that because you also know that there's a specific rewrite that determines which value variables are and aren't relevant.

Otherwise, a generic warning for "unused" variable/value mappings is simply an interface choice that might help inform people of issues elsewhere and/or bad assumptions (e.g. that the resulting graph will depend on certain terms), but that's all.
 
All this relates directly to #85.

PR #24, opened 2021-07-03
Add support for Scan Ops in joint_logprob

kc611:
This PR aims to implements `logprob` for Scan Ops.

```python
    k = at.iscalar("k")
    A = at.vector("A")
    random_dist = at.random.normal(size=10)

    result, updates = aesara.scan(
        fn=lambda prior_result: prior_result + random_dist,
        outputs_info=at.ones_like(A),
        n_steps=k,
    )

    logp = joint_logprob(result, {A: A.type(), k: k.type()})
```

Note : I am building this on top of replacement logic from #19 

***

+ [ ] setting up a general `Scan` replacement framework.


aeppl/opt.py
```diff
@@ -178,8 +158,84 @@ def naive_bcast_rv_lift(fgraph, node):
...
+    # Replace this sub-graph with a `ScanRV`
+    new_node = ScanRV.create_node(node, inner_rvs)
+
+    new_scan_rv = new_node.default_output()
+    new_scan_rv.name = "scan"
+
+    # FIXME: This is pretty hackish
+    fgraph.import_node(new_node, import_missing=True, reason="scan_rv")
+
+    return [new_scan_rv]
```
kc611, line 225: So as my first try I did something similar to what we do with `MixtureRV`. (And, yes, it doesn't work)

This replacement gets stuck in some long 'optimization' computation.

aeppl/opt.py
```diff
@@ -178,8 +158,84 @@ def naive_bcast_rv_lift(fgraph, node):
...
+
+    # GWalk the inner fgraphs for any random variables
+    for node_out in node_outs:
+        out_fgraph = FunctionGraph(
+            outputs=[node_out],
+            clone=True,
+            copy_orphans=False,
+            copy_inputs=False,
+            features=[ShapeFeature()],
+        )
```
kc611, line 206: Is there a way we can get the Scan's inner graph as `FunctionGraph` so we can walk that instead of each output node.

brandonwillard: Are you referring to the output nodes of the inner-graph?

kc611: Yeah like making an entire temporary fgraph just to walk it through to find RV's seems a little bit redundant. Is there a better way to detect the RVs.

aeppl/opt.py
brandonwillard: The replacement performed by `scan_replace` should only occur when the inner-outputs that determine the outer-outputs are all `RandomVariable` nodes (i.e. the (relevant) outputs of scan are effectively a sequence of `RandomVariable` outputs), because that's when we can compute the log-probability with something like `at.sum(scan(lambda ..., value, ...: logprob(inner_out, value), ...))`.

The other rewrites are applied separately to the original `Scan`&mdash;through the normal course of running `logprob_rewrites_db`&mdash;until the inner-graph's `RandomVariable`s are pushed down enough to become the inner and/or outer-outputs.  `scan_replace` is only there to check that this has been done sufficiently enough to replace the `Scan`.

brandonwillard (commented):
We might need to copy-over/create rewrites like `push_out_rvs_from_scan` and `convert_outer_out_to_in` so that the `Scan`s can be further rewritten.  I'm not sure how much of the original `symbolic_pymc` `Scan` code can be used here, but I know that a lot of important steps for building a log-probability graph from a `Scan` are in that old code.

If anything, I should see if I have any illustrations of what these old rewrites do; that would help a lot.

aeppl/opt.py
brandonwillard: `FunctionGraph`s are simple containers that only hold a few `list`s (and sometimes a `dict`), so don't worry too much about creating and discarding them.  The same goes for `ScanArgs`.

aeppl/opt.py
```diff
@@ -185,44 +189,506 @@ def rng_fn(self, rng, *args, **kwargs):
...
 
-    node_outs = node.op.outputs
 
-    if not node_outs:
-        return
+def default_filter_scanargs(x):
+    return x.startswith("inner_") or x.startswith("outer_")
+
+
+class ScanArgs(scan_args):
```
brandonwillard, line 201: This can go in a `utils` module, or, if a lot more code is needed for the `Scan` work, its own module entirely.

Also, we should replace Aesara's current `scan_args` with this `ScanArgs`.

kc611: I'll make a quick PR for `ScanArgs` in Aesara.

kc611: Did it in https://github.com/aesara-devs/aesara/pull/510

brandonwillard:
This PR depends on https://github.com/aesara-devs/aesara/pull/510.

brandonwillard:
I've added some missing tests from `symbolic-pymc`.  Also, I imagine we'll need https://github.com/aesara-devs/aesara/pull/534 to go through before this will work, because the automatic cloning that `ScanArgs` currently performs is not what the borrowed `symbolic-pymc` code expects.

tests/test_scan.py
```diff
@@ -0,0 +1,471 @@
...
+        fn=scan_fn,
+        sequences=[mus_tt, sigmas_tt],
+        non_sequences=[Gamma_rv, rng],
+        outputs_info=[{}],
+        strict=True,
+        name="scan_rv",
+    )
+    Y_rv.name = "Y_rv"
+
+    a = joint_logprob(Y_rv, {})
```
kc611, line 470: This is not being registered as a Scan node.

aeppl/scan.py
```diff
@@ -0,0 +1,327 @@
...
+    If it can, replace it with a `MeasurableScan` that works with `logprob`.
+    """
+
+    if isinstance(node.op, MeasurableScan) or not isinstance(node.op, Scan):
+        return
+
+    # TODO: Determine which outer-outputs are measurable.
+    # Use `curr_scanargs` to connect inner-outputs to outer-outputs.  If we
+    # can determine that an inner-output is a `MeasurableVariable`, then we
+    # can add the corresponding outer-output to `rv_outputs`.
```
brandonwillard, line 285: Without this in place, we're assuming that every `Scan` is a `MeasurableScan` (because we turn every `Scan` into one below), but that's not necessarily true.

We can apply the `Scan` transforms here and then convert the output into a `MeasurableScan`, or we can attempt to determine if a `Scan` can be transformed and do that later.  The former seems easier, though.

brandonwillard: Actually, it looks like we're letting `push_out_rvs_from_scan` run alongside this rewrite, and that should probably produce `MeasurableScan`s.

aeppl/scan.py
```diff
@@ -0,0 +1,332 @@
...
+    This is how we add `rv_sinking_db` and any other log-probability rewrites
+    to the "body" (i.e. inner-graph) of a `Scan` loop.
+    """
+
+    if not isinstance(node.op, Scan):
+        return None
+
+    op = Scan(node.op.inputs, node.op.outputs, node.op.info)
+    new_node = op.make_node(*node.inputs)
+    new_node.op.mode = new_node.op.mode.register([rv_sinking_db])
```
kc611, line 314: Isn't the mode itself `None` over here.

brandonwillard: I just threw that in there to be safe, and it might not even be useful.  There are two `Mode` values attached to the `Op`, the I don't think `mode` is the one that's directly used to compile the inner-graph; however, it may be used in some other way.

aeppl/scan.py
```diff
@@ -0,0 +1,332 @@
...
+    new_inner_out_idx = [
+        curr_scanargs.inner_outputs.index(i) for i in unpushed_inner_rvs
+    ]
+    _ = [
+        outputs.pop(op.var_mappings["outer_out_from_inner_out"][i])
+        for i in new_inner_out_idx
+    ]
+
+    # TODO: Remap any value variables associated with the old `Scan` to this
+    # new `MeasurableScan`?
```
kc611, line 97: I think the `rv_value_remapper` in `joint_logprob` takes care of that no ?

The only issue is it doesn't map variables that are not the `var` (i.e. `S_rv` in our test cases) is that what you meant over here.

aeppl/scan.py
```diff
@@ -0,0 +1,332 @@
...
+            unpushed_inner_rvs.extend(
+                [c for c in clients[n] if c not in curr_scanargs.inner_outputs]
+            )
+
+    if len(unpushed_inner_rvs) != 0:
+        # This means that user used function like
+        # def scan_fn(*args):
+        #     S_t = at.random.categorical()
+        #     Y_t = at.random.normal(arg1[S_t])
+        # return Y_t
```
kc611, line 73: If we dis-allow this kind of functions we wouldn't need the push-out functionality anymore. Right ? Cause all of the outputs are 'required' to be pushed out by the user himself. This function will just act as a check in that case.

kc611:
And the current issue at hand is for some reason, the `rv_remapper` doesn't update variables other than `var` itself. (When running the newly added `test_scan_logprob_basic` the `push_out_rvs` optimization returns mappings for both `S_r` and `Y_rv` but the only one that's updated in this dictionary is `Y_rv`which corresponds to the `var` for which we're getting log probability for)

https://github.com/aesara-devs/aeppl/blob/37a52281bfd9bcad1ec21e21276b9632158e431b/aeppl/joint_logprob.py#L117

Is that intended ?

tests/test_scan.py
```diff
@@ -0,0 +1,589 @@
...
+    # `RandomVariable`
+    with aesara.change_flags(on_unused_input="warn"):
+        res = test_Y_logp.eval(new_test_point)
+
+    exp_res = Y_logp.eval(test_point)
+
+    assert np.array_equal(res, exp_res)
+
+
+def test_scan_logprob_basic():
```
kc611, line 485: Marking for convenience.

tests/test_scan.py
```diff
@@ -0,0 +1,589 @@
...
+    scan_rv = Y_rv.owner
+    assert isinstance(scan_rv.op, Scan)
+
+    m_S_rv = MeasurableScan.create_node(scan_rv, scan_rv.outputs)
+
+    assert m_S_rv.op.inputs == scan_rv.op.inputs
+    assert m_S_rv.op.outputs == scan_rv.op.outputs
+    assert m_S_rv.inputs == scan_rv.inputs
+    # Check clones
+    assert m_S_rv.outputs == scan_rv.outputs
```
brandonwillard: I don't believe this should pass, because the output `TensorVariable`s use identity (e.g. `is`) comparisons.

aeppl/scan.py
brandonwillard: The original "push-out" functionality would automatically add `S_t` as an output to a new, reformulate `Scan`, as I recall; however, in our case, that's not particularly useful, because we need value variables for all the `MeasurableVariable`s in the model, but there isn't a (good) way to specify a value variable for an inner-graph variable like `S_t`.

In other words, yes, I think you have the right idea.

The only way I could see this being done without somehow specifying a value variable for terms like `S_t` is if we never even required value variables.  Instead, these `logprob`-related functions would also return a `dict` mapping `MeasurableVariable`s to value variables that are automatically generated for them by the function.

aeppl/scan.py
brandonwillard: > I think the `rv_remapper` in `joint_logprob` takes care of that no ?

Yes, it generally should.

> The only issue is it doesn't map variables that are not the `var` (i.e. `S_rv` in our test cases). Is that what you meant over here ?

The `PreserveRVMappings` feature might not work for variables that aren't the literal inputs/outputs of a replacement.  I believe that was one of my concerns.  That might not be an issue, though.  We really need to set up some example and work through them to find out.

brandonwillard (commented):
I've pushed a simple fix and test updates.

aeppl/scan.py
brandonwillard: It looks like [the current issue](https://github.com/aesara-devs/aeppl/pull/24/checks?check_run_id=3357658699#step:6:256) actually _is_ due to `PreserveRVMappings` not handling every variable that was replaced.

In this particular case, `PreserveRVMappings.on_change_input` is only ever called with `r` (i.e. the variable being replaced) equal to `Y_rv`.  This is because the `FunctionGraph` that's being used has only one output: `Y_rv`.  The second output of the `Scan` is `S_rv` and the test fails when that output appears in the `joint_logprob` loop.

If `PreserveRVMappings.on_change_input` were called with `r` equal to `S_rv` and `new_r` equal to the corresponding new output produced by `push_out_rvs_from_scan`, this problem wouldn't appear.

We might be able to add all the random variables as outputs to the `fgraph` created in `joint_logprob` and get this to work.  That would guarantee that `PreserveRVMappings.on_change_input` would be called for each one.

brandonwillard: I've just added a fix to that effect.

ricardoV94:
> The only way I could see this being done without somehow specifying a value variable for terms like S_t is if we never even required value variables. Instead, these logprob-related functions would also return a dict mapping MeasurableVariables to value variables that are automatically generated for them when they're encountered/created by these rewrite functions.

Sure, but we need an alternative to specify conditional relationships. We were doing that implicitly depending on whether RVs in the graph were mapped to a value variable or not (e.g., in the Mixture RVs and ChainedTransform / Censored RVs)

aeppl/scan.py
kc611: I think that's how `symbolic-pymc` originally used to calculate the log-probability of Scan ops, no?

https://github.com/pymc-devs/symbolic-pymc/blob/84e8d612c714f502f8d188c1766498f4ff7beecf/symbolic_pymc/theano/pymc3.py#L177

brandonwillard: > I think that's how `symbolic-pymc` originally used to calculate the log-probability of Scan ops, no?

Yeah, as I recall.

brandonwillard:
> > The only way I could see this being done without somehow specifying a value variable for terms like S_t is if we never even required value variables. Instead, these logprob-related functions would also return a dict mapping MeasurableVariables to value variables that are automatically generated for them when they're encountered/created by these rewrite functions.
> 
> Sure, but we need an alternative to specify conditional relationships, when there is potential ambiguity. We were doing that implicitly depending on whether RVs in the graph were mapped to a value variable or not (e.g., in the Mixture RVs and ChainedTransform / Censored RVs)

Yes, that is one of the downsides to the approach.

tests/test_scan.py
```diff
@@ -0,0 +1,518 @@
...
+        res = test_Y_logp.eval(new_test_point)
+
+    exp_res = Y_logp.eval(test_point)
+
+    assert np.array_equal(res, exp_res)
+
+
+def test_scan_logprob():
+    # srng = at.random.RandomStream()
+    rng = aesara.shared(np.random.default_rng(1234))
```
kc611, line 413: Temporarily changed `RandomStreams` to shared `Generators` for the current development run/testing purposes.

aeppl/scan.py
```diff
@@ -0,0 +1,326 @@
...
+
+    output_scan_args.outer_in_seqs += [v[:n_steps] for v in var_slices]
+
+    if not replacing or add_nit_sot:
+        output_scan_args.outer_in_nit_sot += [n_steps]
+
+    if inner_out_fn:
+        output_scan_args.inner_out_nit_sot += [
+            inner_out_fn(
+                input_scan_args, old_inner_out_var, inner_in_var, output_scan_args
```
kc611, line 170: This is where we should be passing the processed value of corresponding RV node using `new_outer_input_vars`. Issue is how do we make sure that each iteration of `Scan` in its `n_step`s of iterations receives the correct value from this array of value.  (i.e. we'll probably need to modify the above `taps` logic for that which will correctly pass the values for each output node as a `seq` type input.)

aeppl/scan.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+def create_inner_out_logp(
+    input_scan_args: ScanArgs,
+    old_inner_out_var: TensorVariable,
+    new_inner_in_var: TensorVariable,
+    output_scan_args: ScanArgs,
+) -> TensorVariable:
+    """Create a log-likelihood inner-output for a `Scan`."""
+    logp = _logprob(
```
kc611, line 207: We'll need a call to `joint_logprob` here rather than `_logprob`.

tests/test_scan.py
brandonwillard: No problem; I don't think there's a strict requirement for either one, but both should be expected to work (although the `assert`s may need to change in response).

brandonwillard (dismissed):
We need to rebase this now that a few of the commits were merged recently.

aeppl/scan.py
brandonwillard: Yeah, there are some design questions/issues with `convert_outer_out_to_in` and multiple value variables.  

I believe `convert_outer_out_to_in` I designed to work with one value variable at a time via the `ScanArgs` (i.e. one would mutate the `ScanArgs` in-place for each random variable).

It should repurpose the original `Scan` so that each new value variable is a new outer-input and&mdash;thus&mdash;has a corresponding inner-input "slice"/"scalar".  All the `taps` logic should be handled by that existing logic, as long as the value variable replacements we use are correct.

These replacements probably need to be the un-`*Subtensor*`ed tensors, and not the ones returned by `aesara.scan`, which cut off the initial values via `*Subtensor*` slices.

brandonwillard:
Here are my old notes on this work from `symbolic-pymc`: [`scan-log-likelihoods.md`](https://gist.github.com/brandonwillard/2740df9b436446f9ce6fd339ee779bb3#file-scan-log-likelihoods-md).  They were written for Theano, but everything still applies to Aesara.

brandonwillard:
I just pushed some changes that should put everything in general working order.

The `convert_outer_out_to_in` tests need to be rewritten so that they test the output of that function directly, and a couple more `Scan` cases would be nice (e.g. using `joint_logprob` on the model in `test_convert_outer_out_to_in_sit_sot`).

brandonwillard:
Let's say we wanted to compute the log-probability for a simple Markov chain like the following:

```python
import numpy as np

import aesara
import aesara.tensor as at

from aeppl.joint_logprob import joint_logprob


aesara.config.on_opt_error = "raise"

S_0_rv = srng.categorical(np.array([0.5, 0.5]), name="S_0")

s_0_vv = S_0_rv.clone()
s_0_vv.name = "s_0"


def step_fn(S_tm1):
    S_t = srng.categorical(np.array([0.5, 0.5]), name="S_t")
    return S_t


S_1T_rv, _ = aesara.scan(
    fn=step_fn,
    outputs_info=[{"initial": S_0_rv, "taps": [-1]}],
    strict=True,
    n_steps=10,
    name="S_0T"
)

S_1T_rv.name = "S_1T"
s_1T_vv = S_1T_rv.clone()
s_1T_vv.name = "s_1T"
```

If we try to construct the log-probability graph, we get an error:

```python
S_0T_logp = joint_logprob({S_1T_rv: s_1T_vv, S_0_rv: s_0_vv})
```

```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/tmp/user/1000/babel-9XUDYv/python-rnH85p in <module>
----> 1 S_0T_logp = joint_logprob({S_1T_rv: s_1T_vv, S_0_rv: s_0_vv})

~/projects/code/python/aeppl/aeppl/joint_logprob.py in joint_logprob(sum, *args, **kwargs)
    218
    219     """
--> 220     logprob = factorized_joint_logprob(*args, **kwargs)
    221     if not logprob:
    222         return None

~/projects/code/python/aeppl/aeppl/joint_logprob.py in factorized_joint_logprob(rv_values, warn_missing_rvs, extra_rewrites, **kwargs)
    157         q_rv_value_vars = [
    158             replacements[q_rv_var]
--> 159             for q_rv_var in outputs
    160             if not getattr(q_rv_var.tag, "ignore_logprob", False)
    161         ]

~/projects/code/python/aeppl/aeppl/joint_logprob.py in <listcomp>(.0)
    158             replacements[q_rv_var]
    159             for q_rv_var in outputs
--> 160             if not getattr(q_rv_var.tag, "ignore_logprob", False)
    161         ]
    162

KeyError: 'for{cpu,scan_fn}.0\n
```

The following two sections summarize the two main features we need to implement before merging.


### Add `MeasurableVariable` detection for `Subtensor` `Scan` output

The first issue is that `S_1T_rv` is actually the output of a `Subtensor`:

```python
print(S_1T_rv.owner)
```

```python
Subtensor{int64::}(for{cpu,S_0T}.0, ScalarConstant{1})
```

In other words, `S_1T_rv = S_0T_rv[1:]` for some `S_0T_rv` that effectively represents `at.concatenate([at.atleast_1d(S_0_rv), S_1T_rv])`.

It is possible to (partially) work around the error above by manually extracting the direct output of the `Scan` node&#x2013;i.e. the output corresponding to `S_0T_rv` (e.g. via `S_1T_rv.owner.inputs[0]`); however, we still need to support the `joint_logprob` arguments/usage demonstrated in the example above, since it's considerably more intuitive. Also, there are some issues with this work-around that will be described below.

We should be able to make some updates to `find_measurable_scans` so that it checks for `Subtensor`s acting on `Scan` outputs and somehow checks these `Subtensor`s against the `Scan`'s TAPS configuration (i.e. attempt to confirm that the `Subtensor` is the output of the `aesara.scan` helper function).

### Support initial values

The second issue is that the initial TAPS values (i.e. the `"initial"` values specified in a call to `aesara.scan`) are not currently included in the resulting log-probability graphs.

If we extract the underlying `S_0T_rv` variables and use that in a call to `joint_logprob`, we can get a log-probability graph:

```python
S_0T_rv = S_1T_rv.owner.inputs[0]

s_0T_vv = S_0T_rv.clone()
s_0T_vv.name = "s_0T"

S_0T_logp = joint_logprob({S_0T_rv: s_0T_vv, S_0_rv: s_0_vv})


aesara.dprint(S_0T_logp, depth=4)
```

```python
Sum{acc_dtype=float64} [id A] ''
 |MakeVector{dtype='float64'} [id B] ''
   |Sum{acc_dtype=float64} [id C] ''
   | |Assert{msg='0 <= p <= 1'} [id D] 's_0_logprob'
   |Sum{acc_dtype=float64} [id E] ''
     |for{cpu,scan_fn} [id F] 's_0T_logprob'

Inner graphs of the scan ops:

for{cpu,scan_fn} [id F] 's_0T_logprob'
 >Assert{msg='0 <= p <= 1'} [id G] 'S_t_vv_logprob'
 > |Elemwise{switch,no_inplace} [id H] ''
 > | |Elemwise{and_,no_inplace} [id I] ''
 > | | |Elemwise{le,no_inplace} [id J] ''
 > | | |Elemwise{lt,no_inplace} [id K] ''
 > | |Elemwise{log,no_inplace} [id L] ''
 > | | |Subtensor{int64} [id M] ''
 > | |TensorConstant{-inf} [id N]
 > |All [id O] ''
 > | |Elemwise{ge,no_inplace} [id P] ''
 > |   |Elemwise{true_div,no_inplace} [id Q] ''
 > |   |InplaceDimShuffle{x} [id R] ''
 > |All [id S] ''
 >   |Elemwise{le,no_inplace} [id T] ''
 >     |Elemwise{true_div,no_inplace} [id Q] ''
 >     |InplaceDimShuffle{x} [id U] ''
```

As this output demonstrates, there is a log-probability term for `S_0_rv` prior (i.e. the sub-graph labeled `"s_0_logprob"`) and another log-probability term for the `Scan`/state sequence&#x2013;labeled `"s_0T_logprob"`&#x2013;that depends on the `s_0_vv` vector; however, the corresponding (outer-)input for the `Scan` doesn't depend on `s_0_vv` in any way.

Let's take a look at the inputs of the model's `Scan` node:

```python
import pprint

from aesara.scan.utils import ScanArgs

## Prevent `dict` sorting
pprint.sorted = lambda arg, *a, **kw: arg

scan_args = ScanArgs.from_node(S_1T_rv.owner.inputs[0].owner)


def labeled_outer_inputs(scan_args):
    """Get ordered, labeled outer inputs for a `Scan`."""
    outer_inputs = [
        "n_steps",
        "outer_in_seqs",
        "outer_in_mit_mot",
        "outer_in_mit_sot",
        "outer_in_sit_sot",
        "outer_in_shared",
        "outer_in_nit_sot",
        "outer_in_non_seqs",
    ]

    return {
        name: getattr(scan_args, name)
        for name in outer_inputs
        if getattr(scan_args, name, None)
    }


model_outer_inputs = labeled_outer_inputs(scan_args)
pprint.pprint(model_outer_inputs)
```

```python
{'n_steps': TensorConstant{10},
 'outer_in_sit_sot': [IncSubtensor{Set;:int64:}.0],
 'outer_in_shared': [RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F9D44833D70>)]}
```

The `'outer_in_sit_sot'` input is the one that corresponds with the state sequence, and it is a vector with its first input set to `s_0_vv`:

```python
aesara.dprint(model_outer_inputs["outer_in_sit_sot"])
```

```python
IncSubtensor{Set;:int64:} [id A] ''
 |AllocEmpty{dtype='int64'} [id B] ''
 | |Elemwise{add,no_inplace} [id C] ''
 |   |TensorConstant{10} [id D]
 |   |Subtensor{int64} [id E] ''
 |     |Shape [id F] ''
 |     | |Rebroadcast{0} [id G] ''
 |     |   |InplaceDimShuffle{x} [id H] ''
 |     |     |categorical_rv{0, (1,), int64, False}.1 [id I] 'S_0'
 |     |       |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F9D44A1B6E0>) [id J]
 |     |       |TensorConstant{[]} [id K]
 |     |       |TensorConstant{4} [id L]
 |     |       |TensorConstant{(2,) of 0.5} [id M]
 |     |ScalarConstant{0} [id N]
 |Rebroadcast{0} [id G] ''
 |ScalarFromTensor [id O] ''
   |Subtensor{int64} [id E] ''
```

Our `convert_outer_out_to_in` needs to identify the original full-sequence variable (i.e. the `AllocEmpty` with ID `B`) and replace it with its value variable.

brandonwillard:
I just added a commit that handles the first remaining issue (i.e. the section in https://github.com/aesara-devs/aeppl/pull/24#issuecomment-922369850 titled "Add MeasurableVariable detection for Subtensor Scan output")

brandonwillard:
I just added the functionality for the last missing feature.  It needs more testing, but at least seems to work for the one simple implemented test case.

brandonwillard:
All right, I'm going to merge this for now; we can put in updates as we come across issues and such.

PR #26, opened 2021-07-07
Implement chained transforms

### LINKED ISSUE #18: Add affine transform support to `total_logprob`
With a new `local_optimizer` addition to `RVSinker`, we could add support for affine/linear transforms of `RandomVariables`.

In other words, the following could be made to work:
```python
import aesara.tensor as at

a, b = at.scalars("ab")
Y_rv = a * at.random.normal(0, 1, name="Y") + b

y_val = Y_rv.type()
logp = joint_logprob(Y_rv, {Y_rv: y_val})
```

ricardoV94:
Closes #18.

TODO:
- [x] Nested transforms for graphs involving a base `RV` and the following ops: `Add`, `Mul`, `Exp`, `Log`
- [x] Deal with base RVs containing size dimensions of `1` and generic loc / shift tensors (see https://github.com/aesara-devs/aeppl/pull/26#issuecomment-900945477)
- [x] ~~Deal properly with broadcasting as discussed in https://github.com/aesara-devs/aeppl/pull/22#issuecomment-875349807~~ deferred to https://github.com/aesara-devs/aeppl/discussions/51
- [x] ~~Deal with discrete variables (either do not allow or do not add jacobian term)~~ Not allowed for now.
- [x] ~~Jacobian of ChainTransform is wonky for non-scalar variables~~ Loc transform was returning a scalar `det`
- [x] Add comprehensive tests
- [x] More transforms can be added later by just expanding the rewrite

### DIFF
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -10,15 +10,23 @@ from aesara.graph.features import AlreadyThere, Feature
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op
 from aesara.graph.opt import GlobalOptimizer, in2out, local_optimizer
+from aesara.scalar import Add, Exp, Log, Mul
 from aesara.tensor.basic_opt import (
     register_specialize,
     register_stabilize,
     register_useless,
 )
+from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.var import TensorVariable
 
-from aeppl.abstract import MeasurableVariable
-from aeppl.logprob import _logprob
+from aeppl.abstract import (
+    MeasurableVariable,
+    _get_measurable_outputs,
+    assign_custom_measurable_outputs,
+)
+from aeppl.logprob import _logprob, logprob
+from aeppl.opt import PreserveRVMappings, measurable_ir_rewrites_db
+from aeppl.utils import walk_model
 
 
 @singledispatch
@@ -203,6 +211,143 @@ class TransformValuesOpt(GlobalOptimizer):
 
     def apply(self, fgraph: FunctionGraph):
         return self.default_transform_opt.optimize(fgraph)
+
+
+class MeasurableTransform(Elemwise):
+    """A placeholder used to specify a log-likelihood for a transformed measurable variable"""
+
+    # Cannot use `transform` as name because it would clash with the property added by
+    # the TransformValuesOpt
+    transform_elemwise: RVTransform
+    measurable_input_idx: int
+
+    def __init__(
+        self, *args, transform: RVTransform, measurable_input_idx: int, **kwargs
+    ):
+        self.transform_elemwise = transform
+        self.measurable_input_idx = measurable_input_idx
+        super().__init__(*args, **kwargs)
+
+
+MeasurableVariable.register(MeasurableTransform)
+
+
+@_get_measurable_outputs.register(MeasurableTransform)
+def _get_measurable_outputs_Transform(op, node):
+    return [node.default_output()]
+
+
+@_logprob.register(MeasurableTransform)
+def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwargs):
+    """Compute the log-probability graph for a `MeasurabeTransform`."""
+    # TODO: Could other rewrites affect the order of inputs?
+    (value,) = values
+    other_inputs = list(inputs)
+    measurable_input = other_inputs.pop(op.measurable_input_idx)
+
+    # The value variable must still be back-transformed to be on the natural support of
+    # the respective measurable input.
+    backward_value = op.transform_elemwise.backward(value, *other_inputs)
+    input_logprob = logprob(measurable_input, backward_value, **kwargs)
+
+    jacobian = op.transform_elemwise.log_jac_det(value, *other_inputs)
+
+    return input_logprob + jacobian
+
+
+@local_optimizer([Elemwise])
+def find_measurable_transforms(
+    fgraph: FunctionGraph, node: Node
+) -> Optional[List[Node]]:
+    """Find measurable transformations from Elemwise operators."""
+    scalar_op = node.op.scalar_op
+    if not isinstance(scalar_op, (Exp, Log, Add, Mul)):
+        return None
+
+    # Node was already converted
+    if isinstance(node.op, MeasurableVariable):
+        return None  # pragma: no cover
+
+    rv_map_feature: PreserveRVMappings = getattr(fgraph, "preserve_rv_mappings", None)
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    # Check that we have a single source of measurement
+    measurable_inputs = [
+        inp
+        for idx, inp in enumerate(node.inputs)
+        if inp.owner
+        and isinstance(inp.owner.op, MeasurableVariable)
+        and inp not in rv_map_feature.rv_values
+    ]
+
+    if len(measurable_inputs) != 1:
+        return None
+
+    measurable_input: TensorVariable = measurable_inputs[0]
+
+    # Do not apply rewrite to discrete variables
+    if measurable_input.type.dtype.startswith("int"):
+        return None
+
+    # Check that other inputs are not potentially measurable, in which case this rewrite
+    # would be invalid
+    other_inputs = tuple(inp for inp in node.inputs if inp is not measurable_input)
+    if any(
+        ancestor_node
+        for ancestor_node in walk_model(
+            other_inputs,
+            walk_past_rvs=False,
+            stop_at_vars=set(rv_map_feature.rv_values),
+        )
+        if (
+            ancestor_node.owner
+            and isinstance(ancestor_node.owner.op, MeasurableVariable)
+            and ancestor_node not in rv_map_feature.rv_values
+        )
+    ):
+        return None
+
+    # Make base_measure outputs unmeasurable
+    # This seems to be the only thing preventing nested rewrites from being erased
+    measurable_input = assign_custom_measurable_outputs(measurable_input.owner)
+
+    measurable_input_idx = 0
+    transform_inputs: Tuple[TensorVariable, ...] = (measurable_input,)
+    transform: RVTransform
+    if isinstance(scalar_op, Exp):
+        transform = ExpTransform()
+    elif isinstance(scalar_op, Log):
+        transform = LogTransform()
+    elif isinstance(scalar_op, Add):
+        transform_inputs = (measurable_input, at.add(*other_inputs))
+        transform = LocTransform(
+            transform_args_fn=lambda *inputs: inputs[-1],
+        )
+    else:
+        transform_inputs = (measurable_input, at.mul(*other_inputs))
+        transform = ScaleTransform(
+            transform_args_fn=lambda *inputs: inputs[-1],
+        )
+
+    transform_op = MeasurableTransform(
+        scalar_op=scalar_op,
+        transform=transform,
+        measurable_input_idx=measurable_input_idx,
+    )
+    transform_out = transform_op.make_node(*transform_inputs).default_output()
+    transform_out.name = node.outputs[0].name
+
+    return [transform_out]
+
+
+measurable_ir_rewrites_db.register(
+    "find_measurable_transforms",
+    find_measurable_transforms,
+    0,
+    "basic",
+    "transform",
+)
 
 
 class LocTransform(RVTransform):
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -205,6 +205,43 @@ class TransformValuesOpt(GlobalOptimizer):
         return self.default_transform_opt.optimize(fgraph)
 
 
+class LocTransform(RVTransform):
+    name = "loc"
+
+    def __init__(self, transform_args_fn):
+        self.transform_args_fn = transform_args_fn
+
+    def forward(self, value, *inputs):
+        loc = self.transform_args_fn(*inputs)
+        return value + loc
+
+    def backward(self, value, *inputs):
+        loc = self.transform_args_fn(*inputs)
+        return value - loc
+
+    def log_jac_det(self, value, *inputs):
+        return at.zeros_like(value)
+
+
+class ScaleTransform(RVTransform):
+    name = "scale"
+
+    def __init__(self, transform_args_fn):
+        self.transform_args_fn = transform_args_fn
+
+    def forward(self, value, *inputs):
+        scale = self.transform_args_fn(*inputs)
+        return value * scale
+
+    def backward(self, value, *inputs):
+        scale = self.transform_args_fn(*inputs)
+        return value / scale
+
+    def log_jac_det(self, value, *inputs):
+        scale = self.transform_args_fn(*inputs)
+        return -at.log(at.abs(scale))
+
+
 class LogTransform(RVTransform):
     name = "log"
 
@@ -216,6 +253,19 @@ class LogTransform(RVTransform):
 
     def log_jac_det(self, value, *inputs):
         return value
+
+
+class ExpTransform(RVTransform):
+    name = "exp"
+
+    def forward(self, value, *inputs):
+        return at.exp(value)
+
+    def backward(self, value, *inputs):
+        return at.log(value)
+
+    def log_jac_det(self, value, *inputs):
+        return -at.log(value)
 
 
 class IntervalTransform(RVTransform):
@@ -322,6 +372,42 @@ class CircularTransform(RVTransform):
 
     def log_jac_det(self, value, *inputs):
         return at.zeros(value.shape)
+
+
+class ChainedTransform(RVTransform):
+    name = "chain"
+
+    def __init__(self, transform_list, base_op):
+        self.transform_list = transform_list
+        self.base_op = base_op
+
+    def forward(self, value, *inputs):
+        for transform in self.transform_list:
+            value = transform.forward(value, *inputs)
+        return value
+
+    def backward(self, value, *inputs):
+        for transform in reversed(self.transform_list):
+            value = transform.backward(value, *inputs)
+        return value
+
+    def log_jac_det(self, value, *inputs):
+        value = at.as_tensor_variable(value)
+        det_list = []
+        ndim0 = value.ndim
+        for transform in reversed(self.transform_list):
+            det_ = transform.log_jac_det(value, *inputs)
+            det_list.append(det_)
+            ndim0 = min(ndim0, det_.ndim)
+            value = transform.backward(value, *inputs)
+        # match the shape of the smallest jacobian_det
+        det = 0.0
+        for det_ in det_list:
+            if det_.ndim > ndim0:
+                det += det_.sum(axis=-1)
+            else:
+                det += det_
+        return det
 
 
 def _create_transformed_rv_op(
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -554,3 +554,152 @@ def test_chained_transform():
         log_jac_det.eval(),
         -np.log(scale) - np.sum(np.log(x_val_forward - loc)),
     )
+
+
+def test_exp_transform_rv():
+    base_rv = at.random.normal(0, 1, size=2, name="base_rv")
+    y_rv = at.exp(base_rv)
+    y_rv.name = "y"
+
+    y_vv = y_rv.clone()
+    logp = joint_logprob({y_rv: y_vv}, sum=False)
+    logp_fn = aesara.function([y_vv], logp)
+
+    y_val = [0.1, 0.3]
+    np.testing.assert_allclose(
+        logp_fn(y_val),
+        sp.stats.lognorm(s=1).logpdf(y_val),
+    )
+
+
+def test_log_transform_rv():
+    base_rv = at.random.lognormal(0, 1, size=2, name="base_rv")
+    y_rv = at.log(base_rv)
+    y_rv.name = "y"
+
+    y_vv = y_rv.clone()
+    logp = joint_logprob({y_rv: y_vv}, sum=False)
+    logp_fn = aesara.function([y_vv], logp)
+
+    y_val = [0.1, 0.3]
+    np.testing.assert_allclose(
+        logp_fn(y_val),
+        sp.stats.norm().logpdf(y_val),
+    )
+
+
+@pytest.mark.parametrize(
+    "rv_size, loc_type",
+    [
+        (None, at.scalar),
+        (2, at.vector),
+        ((2, 1), at.col),
+    ],
+)
+def test_loc_transform_rv(rv_size, loc_type):
+
+    loc = loc_type("loc")
+    y_rv = loc + at.random.normal(0, 1, size=rv_size, name="base_rv")
+    y_rv.name = "y"
+    y_vv = y_rv.clone()
+
+    logp = joint_logprob({y_rv: y_vv}, sum=False)
+    assert_no_rvs(logp)
+    logp_fn = aesara.function([loc, y_vv], logp)
+
+    loc_test_val = np.full(rv_size, 4.0)
+    y_test_val = np.full(rv_size, 1.0)
+
+    np.testing.assert_allclose(
+        logp_fn(loc_test_val, y_test_val),
+        sp.stats.norm(loc_test_val, 1).logpdf(y_test_val),
+    )
+
+
+@pytest.mark.parametrize(
+    "rv_size, scale_type",
+    [
+        (None, at.scalar),
+        (1, at.TensorType("floatX", (True,))),
+        ((2, 3), at.matrix),
+    ],
+)
+def test_scale_transform_rv(rv_size, scale_type):
+
+    scale = scale_type("scale")
+    y_rv = at.random.normal(0, 1, size=rv_size, name="base_rv") * scale
+    y_rv.name = "y"
+    y_vv = y_rv.clone()
+
+    logp = joint_logprob({y_rv: y_vv}, sum=False)
+    assert_no_rvs(logp)
+    logp_fn = aesara.function([scale, y_vv], logp)
+
+    scale_test_val = np.full(rv_size, 4.0)
+    y_test_val = np.full(rv_size, 1.0)
+
+    np.testing.assert_allclose(
+        logp_fn(scale_test_val, y_test_val),
+        sp.stats.norm(0, scale_test_val).logpdf(y_test_val),
+    )
+
+
+def test_transformed_rv_and_value():
+    y_rv = at.random.halfnormal(-1, 1, name="base_rv") + 1
+    y_rv.name = "y"
+    y_vv = y_rv.clone()
+
+    transform_opt = TransformValuesOpt({y_vv: LogTransform()})
+
+    logp = joint_logprob({y_rv: y_vv}, extra_rewrites=transform_opt)
+    assert_no_rvs(logp)
+    logp_fn = aesara.function([y_vv], logp)
+
+    y_test_val = -5
+
+    assert np.isclose(
+        logp_fn(y_test_val),
+        sp.stats.halfnorm(0, 1).logpdf(np.exp(y_test_val)) + y_test_val,
+    )
+
+
+def test_loc_transform_multiple_rvs_fails1():
+    x_rv1 = at.random.normal(name="x_rv1")
+    x_rv2 = at.random.normal(name="x_rv2")
+    y_rv = x_rv1 + x_rv2
+
+    y = y_rv.clone()
+
+    assert joint_logprob({y_rv: y}) is None
+
+
+def test_nested_loc_transform_multiple_rvs_fails2():
+    x_rv1 = at.random.normal(name="x_rv1")
+    x_rv2 = at.cos(at.random.normal(name="x_rv2"))
+    y_rv = x_rv1 + x_rv2
+
+    y = y_rv.clone()
+
+    assert joint_logprob({y_rv: y}) is None
+
+
+def test_discrete_rv_unary_transform_fails():
+    y_rv = at.exp(at.random.poisson(1))
+    assert joint_logprob({y_rv: y_rv.clone()}) is None
+
+
+def test_discrete_rv_multinary_transform_fails():
+    y_rv = 5 + at.random.poisson(1)
+    assert joint_logprob({y_rv: y_rv.clone()}) is None
+
+
+@pytest.mark.xfail(reason="Check not implemented yet, see #51")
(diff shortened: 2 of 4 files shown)


brandonwillard:
> You also mentioned the use of `kanren`

I don't think `kanren` is necessary for this, and it looks like you have a good approach here.  

As long as we're approaching this within the framework of general integral transforms, we should be fine.

ricardoV94:
I added some tests. I think the local rewrite into TransformedRVs cannot work for nested graphs of the kind `Y_rv = a + base_rv * b`. The problem is that we have the `Y_value` value_var at the opposite end of `base_rv`.

```python
base_rv = at.random.normal()
Y_rv = a + base_rv * b
Y_value = Y_rv.type()
```

 Iteratively I would need to do something like:
```python
temp_Y_value, base_rv = Y_value / b, AffineTransform(base_rv, loc=None, scale=b)
Y_rv = a + base_rv
final_Y_value, base_rv = (Y_value - a) / b, AffineTransform(base_rv, loc=a, scale=None)
```
The snippet is a bit messy, but the main point is that I still need a reference to `b` in the last step. I cannot simply use `temp_Y_value` or otherwise I would get `final_Y_value == temp_Y_Value - a == (Y_value / b) - a`.

tests/test_transforms.py
```diff
@@ -187,3 +195,191 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
...
+    y_rv = shift + at.random.poisson(1, name="y_rv")
+    y = y_rv.type()
+    y.name = "y"
+
+    with pytest.raises(NotImplementedError):
+        joint_logprob(y_rv, {y_rv: y})
+
+
+# TODO: shift can broadcast y_rv leading to an incorrect logprob graph
+# which assumes 4 independent draws from y_rv
```
ricardoV94, line 377:
```suggestion
# which assumes more independent draws from y_rv than there are
```

ricardoV94:
After some head-banging I think I made some progress. 

The PR is still very messy, but there is working logic for finding sequences of simple transforms (no reason why it should be limited to the loc / scale cases). It required some duplicated code because the original transforms from PyMC3 go from transformed value vars to untransformed ones, whereas here we need to go the opposite direction. I am still not sure whether it is correct to subtract the log_jac_det or if I should be using the opposite transforms?

Also the loc / scale transforms have one added complexity in that the transformation variables do not correspond to any of the variable inputs (not doing anything clever with loc - scale arguments obviously). My work-around for this was to extend the number of inputs for the Transformed RV, but this feels a bit messy (for instance I cannot easily do a forward pass in a Chained Transform because the relevant arguments at each step are not the outermost ones). I tried returning them with a lambda function of the sort `lambda *inputs: loc`, but this was not working well with multiple transformations.

I'll try to clean the PR a bit better in the next couple of days but feedback is most appreciated.

tests/test_transforms.py
```diff
@@ -172,18 +177,236 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
...
+    assert np.isclose(
+        logp_fn(loc_test_val, scale_test_val, y_test_val),
+        sp.stats.norm(loc_test_val, scale_test_val).logpdf(y_test_val).sum(),
+    )
+
+
+def test_double_scalar_loc_transform():
+    loc1 = at.scalar("loc1")
+    loc2 = at.scalar("loc2")
+    y_rv = loc1 + loc2 + at.random.normal(0, 1, name="y_rv")
```
ricardoV94, line 325: For some weird reason the rewrite fails when the two scalar `loc`s are placed before the `rv`. It complains that I am trying to replace the original scalar `rv` with a vector `value_var`...

It passes with any of the following:
```python
y_rv = loc1 + at.random.normal(0, 1, name="y_rv") + loc2
y_rv = at.random.normal(0, 1, name="y_rv") + loc1 + loc2
```

tests/test_transforms.py
```diff
@@ -172,18 +177,236 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
...
+    loc = at.lscalar("loc")
+    y_rv = loc + at.random.poisson(1, name="y_rv")
+    y = y_rv.type()
+    y.name = "y"
+
+    with pytest.raises(NotImplementedError):
+        joint_logprob(y_rv, {y_rv: y})
+
+
+def test_broadcasted_loc_transform_fails():
```
ricardoV94, line 377: Would a shape assert be a reasonable approach here? This kind of issue crops up in several of the rewrites so it's also fine for me to leave on open issue for this and move on

tests/test_transforms.py
```diff
@@ -172,18 +178,297 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
...
+
+        logp_term_b = sp.stats.norm(a_val).logpdf(b_val)
+
+        total_logp = logp_term_a + log_jacob_term_a + logp_term_b
+        assert np.isclose(logp_vals_fn(transf_a_val, b_val), total_logp)
+
+
+@pytest.mark.parametrize("scalar_loc", (True, False))
+@pytest.mark.parametrize("rv_size", [None, 1, 2, (3, 2)])
+def test_loc_transform(scalar_loc, rv_size):
```
ricardoV94, line 282: These tests fail only for the combination: `scalar_loc == False` and `rv_size==1`.

aeppl/transforms.py
```diff
@@ -220,16 +480,83 @@ def log_jac_det(self, value, *inputs):
...
+        #     inputs = transform.logprob_inputs(*inputs)
+        # return value
+
+    def backward(self, value, *inputs):
+        for transform in reversed(self.transform_list):
+            value = transform.backward(value, *inputs)
+            inputs = transform.filter_non_rv_inputs(*inputs)
+        return value
+
+    def log_jac_det(self, value, *inputs):
```
ricardoV94, line 504: This was copied from PyMC3... but seems to fail for non-scalar base_rvs

tests/test_transforms.py
```diff
@@ -172,18 +170,252 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
...
+    logp = joint_logprob(y_rv, {y_rv: y})
+    logp_fn = aesara.function([b, y], logp)
+
+    b_val = 1.5
+    y_val = [0.1, 0.1]
+
+    assert np.allclose(logp_fn(b_val, y_val), sp.stats.lognorm(b_val).logpdf(y_val))
+
+
+def test_affine_log_transform():
```
ricardoV94, line 378: This tests passes with `size=None`, and it seems related to the ChainedTransform `log_jac_det` returning a wrong value for non-scalar shapes...

tests/test_transforms.py
ricardoV94: The derived TransformedRV has type `TensorType(float64, (True,))`, while the original RV had type `TensorType(float64, vector)`.

ricardoV94: Is there something special about `size=1`?

ricardoV94: Seems weird to have to force the `loc` or `scale` variables to be of the same `TensorType` just for this case.

ricardoV94:
Commented out the invalid broadcast test, so all tests should now pass for the first time!

aeppl/transforms.py
```diff
@@ -67,16 +71,56 @@ def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
         phi_inv = self.backward(value, *inputs)
         return at.log(at.nlinalg.det(at.atleast_2d(jacobian(phi_inv, [value]))))
 
+    def filter_non_rv_inputs(self, *inputs):
+        """
+        Return list of variable inputs that are actually used in the base
+        logprob and logcdf methods.
```
ricardoV94, line 77:
```suggestion
        Return list of variable inputs belong to the base RV
```

aeppl/transforms.py
```diff
@@ -122,6 +174,205 @@ def transform_logprob(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]
...
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return  # noqa
+
+    out_var = node.outputs[0]
+    if out_var not in rv_map_feature.rv_values:
+        return
+
+    implemented_transform_ops = (Add, Mul, Exp, Log)
```
ricardoV94, line 198: `Neg` and `Sub` would be straightforward to add already as well

aeppl/transforms.py
```diff
@@ -65,22 +71,56 @@ def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
...
+
+@_logprob.register(ForwardTransformedRV)
+def forward_transformed_logprob(op, value, *inputs, name=None, **kwargs):
+    """
+    Compute logp graph for a value variable that must still be back-transformed
+    to be on the natural support of the respective random variable.
+    """
+
+    value_rv_support = op.transform.backward(value, *inputs)
+    rv_inputs = op.transform.filter_non_rv_inputs(*inputs)
```
ricardoV94, line 120: Forgot to add this to the original transformed logprob as well.

aeppl/transforms.py
```diff
@@ -133,6 +173,205 @@ def default_transform(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]
...
+        Y = a + at.Normal(0, 1) * b
+    """
+
+    if not isinstance(node.op, Elemwise):
+        return
+
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return  # noqa
```
ricardoV94, line 191: I thought the `# noqa` would make the codecov ignore the non-hit line

aeppl/transforms.py
```diff
@@ -133,6 +173,205 @@ def default_transform(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]
...
+            else:
+                transform_class = ScaleTransform
+                new_input = (
+                    at.mul(other_branches)
+                    if len(other_branches) > 1
+                    else other_branches[0]
+                )
+
+            transform = transform_class(
+                transform_args_fn=lambda *inputs: inputs[-1],
```
ricardoV94, line 287: Not happy with this, I'll try again the `lambda rv: new_input`, but I think it was failing with multiple transforms (overwriting the `new_input`). The indexing works because it's always the same?

aeppl/transforms.py
```diff
@@ -235,9 +487,59 @@ def log_jac_det(self, value, *inputs):
...
+            inputs = transform.filter_non_rv_inputs(*inputs)
+        return value
+
+    def backward(self, value, *inputs):
+        for transform in reversed(self.transform_list):
+            value = transform.backward(value, *inputs)
+            inputs = transform.filter_non_rv_inputs(*inputs)
+        return value
+
+    def log_jac_det(self, value, *inputs):
```
ricardoV94, line 512: This was copied from PyMC3

aeppl/transforms.py
```diff
@@ -235,9 +487,59 @@ def log_jac_det(self, value, *inputs):
...
+            det_ = transform.log_jac_det(value, *inputs)
+            det_list.append(det_)
+            ndim0 = min(ndim0, det_.ndim)
+            value = transform.backward(value, *inputs)
+            inputs = transform.filter_non_rv_inputs(*inputs)
+        # match the shape of the smallest jacobian_det
+        det = 0.0
+        for det_ in det_list:
+            if det_.ndim > ndim0:
+                print("here")
```
ricardoV94, line 526: Oops
```suggestion
```

aeppl/utils.py
```diff
@@ -297,3 +297,81 @@ def get_constant_value(x):
...
+    input_var: TensorVariable,
+    output_var: TensorVariable,
+) -> List[TensorVariable]:
+    """
+    Return ordered list of consecutive variables between input_var and output_var
+    (both included).
+
+    Search is conducted with depth first and stops as soon as the input_var is
+    found for the first time. No checks are done to ensure the found chain is
+    unique (or the shortest possible). A ValueError is raised if the output_var
```
ricardoV94, line 312:
```suggestion
    unique (or the shortest possible). A ValueError is raised if the input_var
```

aeppl/utils.py
```diff
@@ -297,3 +297,81 @@ def get_constant_value(x):
...
+    """
+
+    def expand(node):
+        if node.owner:
+            return node.owner.inputs
+
+    # Generate all nodes up to `input_var` via depth-first-search. We stop iteration
+    # as soon as the `input_var` is found. In the docstrings example:
+    # (
+    #     (y, [abcd, ef])
```
ricardoV94, line 334:
```suggestion
    #     (y, [abcd, ef])  <-- iteration starts here
```

aeppl/transforms.py
```diff
@@ -135,6 +176,205 @@ def default_transform(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]
...
+        if isinstance(transform_node_op, (Add, Mul)):
+            # Find non-rv branches
+            other_branches = [
+                inp
+                for inp in transform_node.inputs
+                if (inp.owner not in rv_chain_nodes and inp is not base_rv)
+            ]
+
+            if isinstance(transform_node_op, Add):
+                transform_class = LocTransform
```
ricardoV94, line 275: Couldn't figure out the right type-hint for `transform_class`

brandonwillard (commented):
Can you add a somewhat high-level description of the approach and design of this feature?

I'm trying to reconcile this with the convolution and scaling rewrites we've been planning, and I want to make sure that we can do both without conflict and/or complications, or&mdash;better yet&mdash;that we can focus on only one approach/framework.

To be clear, I'm talking about `RandomVariable`-specific rewrites like sums of normals and location/scale properties.  They're limited in that they're not as general as a change-of-variables approach, but they have the big advantage that they result in known distributions and they're rather simple to implement.

It seems like we can always add such rewrites, but that they would need to be performed before these transforms.  If that's all there is to it, then I think we're good; if not, then we don't want to dig ourselves into a hole.

Also, we might want to start focusing on those simple rewrites because of their ease and impact, so don't feel rushed to push this through before considering any of those.

ricardoV94:
> Can you add a somewhat high-level description of the approach and design of this feature?

The new local rewrite basically checks if there is a single chain of operations that link an "orphan" value variable to an usused base `RandomVariable`, and replaces the RV graph by a `ChainTransformRV` that contains the equivalent transformations.

```python
y_rv = at.log( at.random.beta(1, 1) * scale) 
# becomes
y_rv = ChainTransformRV(beta(1, 1), [ScaleTransform(scale), LogTransform()])
```

> It seems like we can always add such rewrites, but that they would need to be performed before these transforms. If that's all there is to it, then I think we're good; if not, then we don't want to dig ourselves into a hole.

Yes this should not interfere with that at all. Actually the rewrites can equally easily be applied before or after this optimization, as the `ChainTransformRV` includes all the necessary information to perform the substitution (and we already did the checks to be sure the value var really corresponds to the the base RV and that it has not been used elsewhere in the graph).

```python
y_rv = ChainTransformRV[Normal(0, 1), [ScaleTransform(scale), LocTransform(loc)]] 
# could be rewritten later (or before) as 
y_rv = Normal(0 + loc, 1 * scale)
```


> I'm trying to reconcile this with the convolution and scaling rewrites we've been planning

> To be clear, I'm talking about `RandomVariable`-specific rewrites like sums of normals and location/scale properties. They're limited in that they're not as general as a change-of-variables approach, but they have the big advantage that they result in known distributions and they're rather simple to implement.

The idea would be to apply the loc-scale reparametrization whenever possible as that results in more succinct graphs. Otherwise this rewrite is more general in that it applies to RVs that cannot be parametrized as such.

In addition it is not limited to loc/scale transforms. The reason why I did not limit to this scenario, is that once we have the `ChainedTransformRV` no further complexity was needed to work with other transformations.

By the way, the convolutions are not addressed at all in the local rewrite (and I don't think they should). I actually have an explicit check to make sure the graph does not correspond to a potential convolution.

aeppl/transforms.py
```diff
@@ -200,6 +241,243 @@ def apply(self, fgraph: FunctionGraph):
...
+            )
+        )
+    ]
+    if invalid_ops:
+        return
+
+    # Create list of RV transforms that link the base_rv support to that of the
+    # value var / derived RV. If Y = at.exp(base_rv * 5), this should be
+    # [ScaleTransform(scale=5), ExpTransform()]
+    transform_inputs = list(base_rv.owner.inputs)
```
ricardoV94, line 362:
```suggestion
    transform_inputs = base_rv.owner.inputs
```

ricardoV94:
Okay, so my last roadblock is this edge case:

```python
# loc = at.scalar("loc")  # works
# loc = at.vector("loc")  # fails
loc = at.TensorType("floatX", (True,))("loc")  # works

y_rv = loc + at.random.normal(0, 1, size=1, name="base_rv")
y_rv.name = "y"
y = y_rv.clone()

logp = joint_logprob(y_rv, {y_rv: y})
```
That example only works when `loc` is a `scalar` or the custom `TensorType`. 
Otherwise it complains that I am trying to replaced the `y_rv` by a `LocTransformedRV` of type `TensorType(float64, (True,))`

```python
TypeError: Cannot convert Type TensorType(float64, vector) (of Variable y) into Type TensorType(float64, (True,)). You can try to manually convert y into a TensorType(float64, (True,)).
```

Regardless of the broadcast question in #51 it seems the (1D) `vector` case should work by default no?

After all this is valid:
```python
loc = at.vector("loc")
base_rv = at.random.normal(loc, 1, size=1)
base_rv.eval({loc: [1]})
```

Should this be handled by the new rewrite? Or by the `naive_bcast` rewrite? Or neither?

tests/test_transforms.py
ricardoV94: See https://github.com/aesara-devs/aeppl/pull/26#issuecomment-900945477 for more context

ricardoV94:
My last commit provides a possible solution to the issue described in https://github.com/aesara-devs/aeppl/pull/26#issuecomment-900945477

I am not very happy with it, but couldn't come up with a more satisfying solution yet.

aeppl/transforms.py
```diff
@@ -201,6 +243,255 @@ def apply(self, fgraph: FunctionGraph):
...
+
+    transform_node = transform_op.make_node(*transform_inputs)
+    transform_rv_out = transform_node.default_output()
+    if out_var.name:
+        transform_rv_out.name = out_var.name
+
+    # Tag base_rv as having been used
+    base_rv.tag.ignore_logprob = True
+
+    # Change broadcasting pattern of TransformedRV if needed
```
ricardoV94, line 436: Can I someone recompute the `broadcastable` pattern of the `TransformedRV` to take into consideration the type of the new `loc`/`scale` inputs (and avoid the `MeasurableRebroadcast` altogether)?

brandonwillard (commented):
If I'm correctly understanding the purpose of `MeasurableRebroadcast`, then it would appear that we need another canonicalization that lifts `Rebroadcast` `Op`s.

Can you provide an explicit example in which this issue arises, and show the relevant `aesara.dprint` output?

ricardoV94:
> If I'm correctly understanding the purpose of `MeasurableRebroadcast`, then it would appear that we need another canonicalization that lifts `Rebroadcast` `Op`s.
> 
> Can you provide an explicit example in which this issue arises, and show the relevant `aesara.dprint` output?

It appears in `test_scale_transform` when `scalar_scale = False` and `rv_size=1`

brandonwillard:
> Note that the Rebroadcast is something I had to introduce in the rewrite phase so that I could replace the old generic vector node by what would otherwise be a new length 1 vector TransformedRV.

Broadcasting is just a subset of shape inference and&mdash;just like that&mdash;an `Apply` node's outputs' broadcast patterns are determined by its `Op` and inputs' broadcast patterns.  This means that the mismatched outputs' broadcast patterns can theoretically be fixed by adjusting the broadcast patterns/shapes of their inputs.

This first thing that needs to be done in a situation like this is to determine whether or not the broadcasting inference of the `Op`s involved are working correctly or not.  There have been quite a few instances in which an `Op` has been found to produce incorrect broadcast patterns for its output, and such things need to be addressed directly.

ricardoV94:
Also, for comparison, I set the similar `test_loc_transform` in a way that it passes without the additional Rebroadcast during the rewrite. It requires us to create a `loc` variable with the same broadcast as the RV. The "issue" boils down to:

```python
(at.random.normal(size=1) + at.vector()).type
# TensorType(float64, vector)

(at.random.normal(loc=at.vector(), size=1)).type
# TensorType(float64, (True,))
```

https://github.com/aesara-devs/aeppl/blob/02e6c134a107de8e218d8becf57dba4bc2d56c84/tests/test_transforms.py#L408-L423

brandonwillard:
> Also, for comparison, I set the similar `test_loc_transform` in a way that it passes without the additional Rebroadcast during the rewrite. It requires us to create a `loc` variable with the same broadcast as the RV. The "issue" boils down to:
> 
> ```python
> (at.vector() + at.random.normal(size=1)).type
> # TensorType(float64, vector)
> 
> (at.random.normal(loc=at.vector(), size=1)).type
> # TensorType(float64, (True,))
> ```

Yes, that definitely seems like the/a problem!  We should create an issue for this in Aesara.

brandonwillard:
Actually, the last example is inconsistent; the `loc=at.vector()` and `size=1` are contradictory, since `at.vector()` is a non-broadcastable array&mdash;by definition&mdash;and `size=1` is stating that the output of `at.random.normal` is necessarily broadcastable (i.e. `size` entirely determines the output's shape for scalar `RandomVariable`s).

The second-to-last example makes sense only because the broadcastable `RandomVariable` output (i.e. `size=1`) is broadcasted with the non-broadcastable `at.vector`.

Simply put, in the last example, `size` shouldn't be `1`.

@ricardoV94, we need to address whatever it is that's giving rise to that situation.

ricardoV94:
> @ricardoV94, we need to address whatever it is that's giving rise to that situation.

So the question in this PR is whether we want to support the derivation for this kind of graph:

```python
loc = at.vector()
rv = at.random.normal(0, 1, size=1) + loc
vv = rv.type()
```
In theory `loc` could be of length 1 at evaluation time, which would define a "valid" RV graph. 

If we want to support this, we would need to somehow make the derived rv have a vector type (one way could be through the `MeasurableBroadcast` I added),

If we don't want to support this, then there is no further roadblock to this PR, I would just revert the MeasurableBroadcast commit and let the rewrites fail naturally for this type of graph.

Users can still make this work by providing a valid `loc`, via `loc = at.TensorType("floatX", (True,))("loc")`

brandonwillard:
> So the question in this PR is whether we want to support the derivation for this kind of graph:

No, the question is "Why are we getting inconsistent `loc` and `size` values?".  [As I stated above](https://github.com/aesara-devs/aeppl/pull/26#issuecomment-921384263), `loc=at.vector()` and `size=1` are not consistent parameters, because the former implies that the output should have `broadcastable == (False,)` and the latter implies it should have `broadcastable == (True,)`.

ricardoV94:
Alright, I think I am gonna leave the graph `at.vector('x') + at.random.normal(size=1, name="base_rv")` not supported for the time being.

aeppl/opt.py
```diff
@@ -191,6 +192,9 @@ def query(self, *tags, **kwtags):
 rv_sinking_db.register("incsubtensor_lift", incsubtensor_rv_replace, -5, "basic")
 
 logprob_rewrites_db.register("sinking", rv_sinking_db, -10, "basic")
+logprob_rewrites_db.register(
```
ricardoV94, line 195: move this to transforms.py file

aeppl/transforms.py
```diff
@@ -80,16 +100,39 @@ def transformed_logprob(op, values, *inputs, use_jacobian=True, **kwargs):
...
         original_forward_value = op.transform.forward(value, *inputs)
         jacobian = op.transform.log_jac_det(original_forward_value, *inputs)
+        jacobian.name = "logprob_jac"
         logprob += jacobian
 
     return logprob
 
 
+@_logprob.register(ForwardTransformedRV)
+def forward_transformed_logprob(op, values, *inputs, **kwargs):
```
ricardoV94, line 117: Add jacobian kwarg?

tests/test_transforms.py
```diff
@@ -360,3 +365,379 @@ def test_TransformValuesMapping():
...
+    ab = a + b
+    ac = a + c
+    x = ab + ac
+    fgraph = FunctionGraph(outputs=[x], clone=False)
+
+    assert not _get_single_vars_between_in_out(a, x, fgraph)
+    assert tuple(_get_single_vars_between_in_out(b, x, fgraph)) == (b, ab, x)
+    assert tuple(_get_single_vars_between_in_out(c, x, fgraph)) == (c, ac, x)
+
+    # Graph where input is not related to requseted output
```
ricardoV94, line 734:
```suggestion
    # Graph where input is not related to requested output
```

ricardoV94:
This one is ready for review

ricardoV94:
This one is ready for review. 

I reverted the Rebroadcast for now. In a subsequent PR we could add a rewrite that "lifts" a Rebroadcast to the RV inputs or changes the RV size, so that we can output rebroadcasted MeasureVariables from our rewrites (or apply it directly in the rewrites)

brandonwillard (commented):
Looks like this needs to be rebased.

aeppl/transforms.py
```diff
@@ -196,6 +240,248 @@ def apply(self, fgraph: FunctionGraph):
...
+    ef = e + f
+    y = abcd + ef
+
+    get_single_vars_between_input_output(c, y) -> [(c, cd, abcd, y)]
+
+    """
+
+    def expand(node):
+        clients = fgraph.clients[node]
+        if len(clients) != 1 or clients[0][0] == "output":
```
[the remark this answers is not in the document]
ricardoV94: This also avoids "diamond" structures like:

```
x───►g(x)───► f(g(x),h(x))
│             ▲
└───►h(x)─────┘
```

ricardoV94:
With #129 this PR is now much more straightforward! Ready for review

ricardoV94:
Hmm, does the IR "reverter" work in this case where nodes are removed instead of replaced?

aeppl/transforms.py
```diff
@@ -205,6 +213,181 @@ def apply(self, fgraph: FunctionGraph):
...
+            ancestor_node.owner
+            and isinstance(ancestor_node.owner.op, MeasurableVariable)
+            and ancestor_node not in rv_map_feature.rv_values
+        )
+    ):
+        return None
+
+    # Make base_measure outputs unmeasurable
+    # This seems to be the only thing preventing nested rewrites from being erased
+    measurable_input = assign_custom_measurable_outputs(measurable_input.owner)
```
ricardoV94, line 314: @brandonwillard does this manipulation make sense now that rewrites apply regardless of being directly valued? I had to introduce it so that nested rewrites would work, otherwise the intermediate nodes would be replaced back to the original form. Clip, Cumsum, and Mixture all do this as well.

ricardoV94:
I had to change strategy away from the transform machinery, because I couldn't see a way to incorporate inputs that are not part of the base measurable variable, which is needed for the `add` and `mul` cases.

aeppl/transforms.py
brandonwillard: Can you perform a check in `PreserveRVMappings` that prevents the relevant variables from being reverted when these transforms are performed?

aeppl/transforms.py
```diff
@@ -205,6 +213,180 @@ def apply(self, fgraph: FunctionGraph):
         return self.default_transform_opt.optimize(fgraph)
 
 
+class MeasurableTransform(Elemwise):
+    """A placeholder used to specify a log-likelihood for a transformed RV"""
+
+    # Cannot use `transform` as name because it would clast with the property added by
```
brandonwillard, line 219:
```suggestion
    # Cannot use `transform` as name because it would clash with the property added by
```

aeppl/transforms.py
```diff
@@ -205,6 +213,180 @@ def apply(self, fgraph: FunctionGraph):
         return self.default_transform_opt.optimize(fgraph)
 
 
+class MeasurableTransform(Elemwise):
+    """A placeholder used to specify a log-likelihood for a transformed RV"""
```
brandonwillard, line 217: Do you mean "measurable variable" when you say "RV"?  We need to sort out our terminology, because we tend to use "measurable variable", "random variable", and `RandomVariable` interchangeably, but there are some important and subtle differences.

aeppl/transforms.py
```diff
@@ -205,6 +213,180 @@ def apply(self, fgraph: FunctionGraph):
...
+    elif isinstance(scalar_op, Add):
+        transform_inputs = (measurable_input, at.add(*other_inputs))
+        transform = LocTransform(
+            transform_args_fn=lambda *inputs: inputs[-1],
+        )
+    else:
+        transform_inputs = (measurable_input, at.mul(*other_inputs))
+        transform = ScaleTransform(
+            transform_args_fn=lambda *inputs: inputs[-1],
+        )
```
brandonwillard, line 331: This seems like very dispatch-worthy logic.

aeppl/transforms.py
ricardoV94: Not sure what such check would look like. How would we know if a variable in `PreserveRVMappings.measurable_conversions` is there because it was ultimately needed for deriving a valued variable or it was not needed.

aeppl/transforms.py
ricardoV94: Measurable variable is correct here, although our measurable variables are ultimately RandomVariables, just not atomic ones.

aeppl/transforms.py
ricardoV94: Yeah, I tried to make it more agnostic than what is actually needed for the current rewrites but I am still not sure what information will be needed going forward. Now I realize that for non-commutative Ops we will also need to know which of the inputs is measurable. There might be other details I am missing. Shall we leave it as is reassess when we expand this with more transforms?

aeppl/transforms.py
brandonwillard: > Measurable variable is correct here, although our measurable variables are ultimately RandomVariables, just not atomic ones.

It should be the other way around in our typing: i.e. `issubclass(RandomVariable, MeasureVariable) is True`.

ricardoV94: And it is. I was thinking in terms of random generation behavior, in the sense that (5 + RV) is an RV

aeppl/transforms.py
brandonwillard: Yeah, it's fine as is for now.

brandonwillard (approved):

PR #30, opened 2021-07-21
Transform fix

ricardoV94:
This PR fixes an inconsistency in the `transformed_logprob`. The jacobian should be computed from the original forward transformed value, but due to the synchronous replacements in `joint_logprob` it was being computed from the backward transformed value.

On PyMC3 V3, these lines show that the jacobian is computed on the original transformed value, whereas the logp is computed on the untransformed value: https://github.com/pymc-devs/pymc3/blob/10c914b92c649b018be09e6dab5a5bc905d4f41e/pymc3/distributions/transforms.py#L177-L197

~~I still have to check if V4 is handling this well.~~ In V4, it's working fine because the original `value_vars` are passed directly to the jacobian method when collecting the logp terms, and only the `rvs` are subsequently replaced by the back-transformed `value_vars`:

https://github.com/pymc-devs/pymc3/blob/bb253c5f464b80c14b6a52529c26b5e72358b879/pymc3/distributions/logprob.py#L218

I added a simple numerical test that passes in this PR but fails in main. 

Also added a `log_jac_det` method to the `logodds` transform and a test for the fallback method in the super `RVTransform` class

### DIFF
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -192,6 +192,10 @@ class LogOddsTransform(RVTransform):
     def forward(self, value, *inputs):
         return at.log(value / (1 - value))
 
+    def log_jac_det(self, value, *inputs):
+        sigmoid_value = at.sigmoid(value)
+        return at.log(sigmoid_value) + at.log1p(-sigmoid_value)
+
 
 class StickBreaking(RVTransform):
     name = "stickbreaking"
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -13,11 +13,10 @@ from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.var import TensorVariable
 
 from aeppl.logprob import _logprob
-from aeppl.utils import logsumexp
 
 
 @singledispatch
-def _transformed_rv(
+def _default_transformed_rv(
     op: Op,
     node: Node,
 ) -> Optional[TensorVariable]:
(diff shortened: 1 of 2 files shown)


aeppl/transforms.py
```diff
@@ -73,11 +71,16 @@ class TransformedRV(RandomVariable, metaclass=DistributionMeta):
...
+    """
+    Compute logp graph for a value variable that was back-transformed to be on
+    the natural support of the respective random variable.
+    """
 
     logprob = _logprob(op.base_op, value, *inputs, name=name, **kwargs)
 
-    jacobian = op.transform.log_jac_det(value, *inputs)
+    original_forward_value = op.transform.forward(value, *inputs)
+    jacobian = op.transform.log_jac_det(original_forward_value, *inputs)
```
ricardoV94, line 83: These two lines are the bugfix, everything else is just docstrings / tests

brandonwillard (approved):
Try to keep independent changes in their own commits (e.g. the refactoring/renaming); we want to be able to revert things without losing other non-required changes&mdash;for example.  Also, it's much easier to review and understand all the changes that way.

Other than that, this looks fine.

ricardoV94:
> Try to keep independent changes in their own commits (e.g. the refactoring/renaming); we want to be able to revert things without losing other non-required changes&mdash;for example.  Also, it's much easier to review and understand all the changes that way.
> 
> Other than that, this looks fine.

Good point

PR #35, opened 2021-07-27
Use Interval default transform in `HalfNormal`

### LINKED ISSUE #31: HalfNormal should have interval transform
The `HalfNormal` in aeppl seems to accomodate the `loc` parameter, which means the default transform should be `Interval[loc, None]`, assuming the loc is just shifting the cutoff point of the `HalfNormal` from zero.

ricardoV94:
Closes #31

### DIFF
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -260,19 +260,23 @@ def create_default_transformed_rv_op(
 
 TransformedUniformRV = create_default_transformed_rv_op(
     at.random.uniform,
-    IntervalTransform(lambda *inputs: inputs[3:]),
+    # inputs[3] = lower; inputs[4] = upper
+    IntervalTransform(lambda *inputs: (inputs[3], inputs[4])),
 )
 TransformedParetoRV = create_default_transformed_rv_op(
     at.random.pareto,
+    # inputs[3] = alpha
     IntervalTransform(lambda *inputs: (inputs[3], None)),
 )
 TransformedTriangularRV = create_default_transformed_rv_op(
     at.random.triangular,
+    # inputs[3] = lower; inputs[5] = upper
     IntervalTransform(lambda *inputs: (inputs[3], inputs[5])),
 )
 TransformedHalfNormalRV = create_default_transformed_rv_op(
     at.random.halfnormal,
-    LogTransform(),
+    # inputs[3] = loc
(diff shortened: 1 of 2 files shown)


twiecki (approved):

aeppl/transforms.py
```diff
@@ -272,7 +272,7 @@ def create_default_transformed_rv_op(
 )
 TransformedHalfNormalRV = create_default_transformed_rv_op(
     at.random.halfnormal,
-    LogTransform(),
+    IntervalTransform(lambda *inputs: (inputs[3], None)),
```
twiecki, line 275: I find this way of indexing into `input` very hard to read.

ricardoV94: Would `IntervalTransform(lambda *_, loc, sigma: (loc, None))` be better?

twiecki: I think better would be if we unpacked them `rv, op, ... = inputs`. Or turn inputs into a dict or NamedTuple.

ricardoV94: The first 3 inputs are always the same, it would be `rng, dtype, size, inp1, inp2, inpN = inputs` but that adds a lot of noise I think. I don't see where the named tuples could be introduced.

Alternatively we could pass the argument indexes ignoring the first three:

`IntervalTransform(lower_arg_idx=0, upper_arg_idx=None)`, which is perhaps easier to read than having to subtract 3 to know which actual "input" it is.

twiecki: Why not do `rng, dtype, size, *inp = inputs`?

ricardoV94: Seems strange to type out explicitly only those variables that we will not use

twiecki: I like `IntervalTransform(lambda *_, loc, sigma: (loc, None))`.

ricardoV94: Actually that's not allowed, `loc` and `sigma` would become keyword only arguments...

twiecki: OK well maybe just leave it and add a comment. Or go with your option without the lambda.

ricardoV94: Went with a comment

twiecki (approved):

aeppl/transforms.py
twiecki: Looks good!

PR #36, opened 2021-07-28
Fix nested transformations

### LINKED ISSUE #29: How to generate graph with nested transforms?
I don't understand how to generate a graph with multiple automatic transforms:

```python
import aesara
import aesara.tensor as at
from aesara.graph.opt import in2out, out2in

from aeppl.transforms import transform_logprob

# Does not matter if I set ignore_newtrees to True, or if I use out2in
transform_opt = in2out(transform_logprob, ignore_newtrees=False)

l_rv = at.random.uniform(0, 1, name='l_rv')
u_rv = at.random.uniform(2, 4, name='u_rv')
# l_rv = at.random.normal(name='l_rv')
# u_rv = at.random.normal(name='u_rv')
y_rv = at.random.uniform(l_rv, u_rv, name='y_rv')

l = l_rv.type()
l.name = 'l'
u = u_rv.type()
u.name = 'u'
y = y_rv.type()
y.name='y'

logp = joint_logprob(y_rv, {y_rv:y, l_rv:l, u_rv:u}, extra_rewrites=transform_opt)
logp.eval({l: -1, u:1, y:0}) 
```

```python
AttributeError: 'numpy.random._generator.Generator' object has no attribute 'uniform_interval'
```

I changed this line so that the name would not be a tuple:
https://github.com/aesara-devs/aeppl/blob/b179a6e1036b63179800bed2a17e51cc46e9b5b1/aeppl/transforms.py#L232
to `cls_dict["name"] = f"{rv_name}_{trans_name}"`

But otherwise the error comes from the same source, i.e., trying to access the `rng_fn` of the `uniform_interval`

ricardoV94:
There was an issue in hierarchical models with transformed variables, **when** the the transformation of downstream value variables depended on other RVs (as happens in the `Interval` transform). We need to also replace the RVs in the `value` graph.

Fixes #29

### DIFF
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -152,8 +152,8 @@ def joint_logprob(
                 # Replace `RandomVariable`s in the inputs with value variables.
                 # Also, store the results in the `replacements` map so that we
                 # don't need to redo these replacements.
-                value_var_inputs, _ = rvs_to_value_vars(
-                    node.inputs,
+                (q_rv_value_var, *value_var_inputs), _ = rvs_to_value_vars(
+                    [q_rv_value_var, *node.inputs],
                     initial_replacements=replacements,
                 )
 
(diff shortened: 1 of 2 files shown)


brandonwillard (approved):

brandonwillard:
Good catch!

PR #39, opened 2021-07-29
Allow variable-specific transforms

### LINKED ISSUE #32: Make transforms more flexible
The current transforms rewrite looks for all RVs that have a standard TransformedRV and automatically converts those (and the respective value variables). To plugin `aeppl` into PyMC3 in the future, we probably need something more flexible where we specify **what** transforms we want to apply to **which** value variables.

This should also allow us to apply transforms to value variables that are associated with dynamically derived RVs.

For example:
```python
import aesara.tensor as at

shift_rv = at.random.normal(0, 1)
y_rv = at.random.exponential(1) + shift

shift = shift_rv.clone()
y = y_rv.clone()

logp = joint_logprob(y_rv, {y_rv: y, shift_rv: shift}, transfoms={y: Interval(shift_rv, None)})
```
I think the logic could still be implemented as a custom "rewrite", I just used a new keyword to simplify the example.

ricardoV94:
Closes #32

I implemented an opt container `TransformValuesOpt` that allows one to specify what transforms to apply to different value variables. This example illustrates the interface:

```python
import aesara.tensor as at
from aeppl.transforms import DEFAULT_TRANSFORM, LogOddsTransform, TransformValuesOpt

x_rv = at.random.uniform(0, 1, name="x")
y_rv = at.random.uniform(x_rv, 5, name="y")
z_rv = at.random.uniform(y_rv, 10, name="z")

x = x_rv.clone()
y = y_rv.clone()
z = z_rv.clone()

transform_opt = TransformValuesOpt(
    {
        x: LogOddsTransform(),
        y: DEFAULT_TRANSFORM,
        z: None,
    }
)

joint_logprob(
    z_rv,
    {x_rv: x, y_rv: y, z_rv: z},
    extra_rewrites=transform_opt
)
```


ricardoV94:
Apologies for the large shuffling around and renaming of objects. The file was getting quite hard navigate. Hopefully it is now a bit more readable.

ricardoV94:
Okay, I split the PR into two commits. The first commit is only a refactoring of `transforms.py` and does not change any functionality. The second commit adds the new functionality.

aeppl/transforms.py
```diff
@@ -47,50 +51,57 @@ def class_transformed_rv(op, node):
...
-        jacobian.name = f"{name}_logprob_jac"
+    new_op_type = type(f"Transformed{rv_type_name}", (TransformedRV,), cls_dict)
 
-    return logprob + jacobian
+    return new_op_type
+
+
+# =============
+# Transform Opt
+# =============
```
brandonwillard, line 104: Comment-based sectional delimiters are just a personal style choice, and not something that particularly helps the readability/navigation of the code, so let's not add these sorts of things.

aeppl/transforms.py
```diff
@@ -56,25 +54,32 @@ class TransformedRV(RandomVariable, metaclass=TransformedRVMeta):
 
 
 @singledispatch
-def _default_transformed_rv(
+def _get_default_transformed_rv_op(
```
brandonwillard: This looks like a purely refactoring change.  Is it necessary for the functionality changes implied by this commit?  If so, you'll need to clarify; otherwise, it would be better in a separate commit, because then we can focus on the functionality itself.

aeppl/transforms.py
```diff
@@ -285,67 +365,82 @@ def log_jac_det(self, value, *inputs):
 # ======================
 
 
-TransformedUniformRV = create_default_transformed_rv_op(
+TransformedUniformRV = create_transformed_rv_op(
```
brandonwillard, line 368: These renamings appear to be undoing renamings from an earlier commit of yours: 52dd503.  Is this intentional, or are these changes unintended regressions that snuck into the commit?

brandonwillard (dismissed):
There are a lot of style and refactoring changes that don't appear to be necessary and distract from the functional changes (both within and across commits), but the main idea/approach (i.e. `TransformValuesMapping`) seems fine.

I don't know if we need the `GlobalOptimizer`-like `TransformValuesOpt` class to wrap the use of `TransformValuesMapping`, though.

If you put the unrelated refactoring and style changes into separate commits, we can review and consider those, but that's going to slow the acceptance of the important functional changes that define _this_ PR, so it would be much better to put those in a separate PR that we can review independently.

aeppl/transforms.py
```diff
@@ -66,76 +34,231 @@ def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
...
+# =============
+# Transform Opt
+# =============
+
+
+DEFAULT_TRANSFORM = object()
+
+
+class TransformValuesMapping(Feature):
+    r"""Store dictionary indicating the transforms to apply to each value variable"""
```
brandonwillard, line 111:
```suggestion
    r"""A `Feature` that maintains a map between value variables and their transforms."""
```

aeppl/transforms.py
```diff
@@ -66,76 +34,231 @@ def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
...
+
+    def __init__(self, transform_values):
+        self.transform_values = transform_values
```
brandonwillard, line 114: It would make more sense to call this variable something like `values_to_transforms`, or `vars_to_transforms`.  The current name implies somethings else.

aeppl/transforms.py
ricardoV94: The returned object of that method is no longer a node but the `TransformedRV` `Op`. The reason for that is that I have to create the node for non default transforms anyway in the local optimizer, so there was no reason to specialize the dispatch method as much

aeppl/transforms.py
ricardoV94: Intentional, the previous change made the name better reflect was the method was doing at the time of the previous commit, and the same applies here.

ricardoV94: They are not an exact revert. It has the new `op` suffix

aeppl/transforms.py
ricardoV94: I like `values_to_transforms`

ricardoV94:
> I don't know if we need the `GlobalOptimizer`-like `TransformValuesOpt` class to wrap the use of `TransformValuesMapping`, though.

My initial approach involved passing the feature and the rewrite as a dictionary but it made the code a bit more messy in the `joint_logprob`.

Another option would be to have an `extra_ features` like `extra_rewrites` which is called before, but I am not sure that looks better...

aeppl/transforms.py
brandonwillard: We definitely don't want to start basing our object names so directly on their exact input and return types.  That information is suffiently conveyed by the type information and docstrings; the names really only need to reflect the basic actions/operations, and some ambiguity/inaccuracy is fine as long as it isn't outright contradictory.

ricardoV94: I am not usually this nitpick,l with the names.

It is just that in this module there are so many things called `transform*` that it becomes hard to keep track. 

In any case, I am fine with reverting this one. I don't like the long name either.

ricardoV94: On the pro side, it aligns with the `create_transformed_rv_op` name, and those are returning the same type of object now

aeppl/transforms.py
ricardoV94: I'll remove them

aeppl/transforms.py
```diff
@@ -47,50 +51,52 @@ def class_transformed_rv(op, node):
...
 
-    def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
-        """Construct the log of the absolute value of the Jacobian determinant."""
-        # jac = at.reshape(
-        #     gradient(at.sum(self.backward(value, *inputs)), [value]), value.shape
-        # )
-        # return at.log(at.abs_(jac))
-        phi_inv = self.backward(value, *inputs)
-        return at.log(at.nlinalg.det(at.atleast_2d(jacobian(phi_inv, [value]))))
+@singledispatch
```
ricardoV94, line 56: This was moved here, so that in the next commit I can change the return type to `Optional[TransformedRV]` without additional changes

aeppl/transforms.py
```diff
@@ -56,25 +54,32 @@ class TransformedRV(RandomVariable, metaclass=TransformedRVMeta):
 
 
 @singledispatch
-def _default_transformed_rv(
+def _default_transformed_rv_op(
```
ricardoV94, line 57: I still think it makes sense to add the `_op` suffix in this commit, as it exists in the related `create_transformed_rv_op`. 

I imagine the original name distinction was exactly because of the different returned types, and that if we were implementing this from scratch we would have chosen similar signatures...

aeppl/transforms.py
```diff
@@ -66,76 +34,218 @@ def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
...
+class TransformValuesMapping(Feature):
+    r"""A `Feature` that maintains a map between value variables and their transforms."""
+
+    def __init__(self, values_to_transforms):
+        self.values_to_transforms = values_to_transforms
+
+    def on_attach(self, fgraph):
+        if hasattr(fgraph, "transform_values_mapping"):
+            raise ValueError(
+                f"{fgraph} already has the `TransformValuesMapping` feature attached."
```
ricardoV94, line 114: Is it worth it to add a test for this `ValueError`? There is something similar in the `PreserveRVMappings` that is also not being tested AFAIK. 

Also, on the Aesara docs, it suggests we should raise the more specific `AlreadyThere` exception: https://aesara.readthedocs.io/en/latest/library/graph/fgraph.html?highlight=feature#aesara.graph.features.Feature.on_attach

Should we do that?

brandonwillard: Yeah, the `AlreadyThere` approach sounds better (assuming it really was/is fully supported).

brandonwillard:
I've simplified this commit and added a missing test for graphs containing multi-output `Op`s that do not also have default outputs (e.g. `SVD`).

Right now, my only real concern with this implementation is that it requires one to specify that a default transform should be used.  Instead, it should probably apply the default transforms without the need to specify them.

rlouf:
On this last point: should the decision of what constitutes a default transform and whether to apply it be made at the level of this library? It feels closer to the application than what I thought `aeppl` was going to be.

My inclination would be to apply no transformation if nothing is specified.

aeppl/transforms.py
```diff
@@ -135,6 +158,63 @@ def default_transform(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]
...
+            a specific value variable it will not be transformed.
+
+        Raises
+        ======
+        RuntimeError
+            If any of the specified value variables is not successfully transformed
+
+        """
+
+        self.values_to_transforms = values_to_transforms.copy()
```
ricardoV94, line 201: I was copying the dict when passing it to the feature so that the optimizer could be reused without having to reinitialize it (otherwise the class copy of the dict is spent inplace).

I feel that is less error prone (eg if in an iterative environment like Jupyter one makes the opt in a separate cell and runs the joint logprob twice the second one would silently not do anything

aeppl/transforms.py
```diff
@@ -89,6 +91,13 @@ def transformed_logprob(op, value, *inputs, name=None, **kwargs):
...
 
+class DefaultTransformSentinel:
+    pass
+
+
+DEFAULT_TRANSFORM = DefaultTransformSentinel()
+
+
 @local_optimizer(tracks=None)
 def default_transform(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]:
```
ricardoV94, line 102: Calling this opt `default_transform` feels wrong.

aeppl/transforms.py
```diff
@@ -135,6 +158,63 @@ def default_transform(fgraph: FunctionGraph, node: Node) -> Optional[List[Node]]
...
+        ==========
+        values_to_transforms
+            Mapping between value variables and their transformations.
+            Each value variable can be assigned one of `RVTransform`,
+            ``DEFAULT_TRANSFORM``, or ``None``. If a transform is not specified for
+            a specific value variable it will not be transformed.
+
+        Raises
+        ======
+        RuntimeError
```
ricardoV94, line 196: It's just a warning now

aeppl/transforms.py
ricardoV94: Also don't you prefer the more specific return type hint?

ricardoV94:
> Right now, my only real concern with this implementation is that it requires one to specify that a default transform should be used.  Instead, it should probably apply the default transforms without the need to specify them.

I'd rather require users to be explicit. Transforms already have a tendency to be very invisible / surprising. For other libraries like PyMC it won't matter, as the dict will be created programatically.

We could add a less flexible opt that Only applies default transforms (just inherit from this and use the value_vars in the fgraph)

ricardoV94:
> On this last point: should the decision of what constitutes a default transform and whether to apply it be made at the level of this library? It feels closer to the application than what I thought `aeppl` was going to be.

The defaults can be overwritten easily by another user / library. Otherwise I think they are pretty standard defaults


> My inclination would be to apply no transformation if nothing is specified.

Me too

aeppl/transforms.py
ricardoV94: A test should be added if we want to support this functionality

brandonwillard:
> On this last point: should the decision of what constitutes a default transform and whether to apply it be made at the level of this library? It feels closer to the application than what I thought `aeppl` was going to be.
> 
> My inclination would be to apply no transformation if nothing is specified.

When we say "default" we mean "the default for the transformation rewrites feature".  In AePPL, transformations aren't applied unless the user explicitly adds the transformation rewrites (see the unit tests for examples).

Otherwise, regarding the default transforms performed by the transform rewriter: those are configurable by customizing a generic/single dispatch function.

aeppl/transforms.py
ricardoV94: That is `Optional[List[TransformedRV]]`

aeppl/transforms.py
brandonwillard: Yes, by design; it's not clear that an error is more suitable.

ricardoV94: I meant that the docstrings are incorrect now

aeppl/transforms.py
brandonwillard: The only logical reason for this step was the check being performed in the original code and its accompanying error, and, since the latter was already questionable (and now just a warning), the `list.pop` and `copy` steps should be removed entirely.

aeppl/transforms.py
brandonwillard: These are unrelated changes, but we should create some PRs for them.

brandonwillard:
> I'd rather require users to be explicit. Transforms already have a tendency to be very invisible / surprising. For other libraries like PyMC it won't matter, as the dict will be created programatically.

OK, we'll go with that.

aeppl/transforms.py
```diff
@@ -240,8 +320,24 @@ def log_jac_det(self, value, *inputs):
 def create_default_transformed_rv_op(
     rv_op: Op,
     transform: RVTransform,
+    *,
+    default: bool = True,
```
ricardoV94, line 324: I feel the default should be there option with least side effects (and these dispatch side effects are quite significant imho). Even though it definitely makes things more verbose for what is the most common use.

aeppl/transforms.py
ricardoV94: Done: https://github.com/aesara-devs/aeppl/issues/43

ricardoV94:
This seems ready, only thing left to do is deciding on whether the `default` of `create_default_transformed_rv_op` should be `True` or `False`.

aeppl/transforms.py
```diff
@@ -89,52 +90,113 @@ def transformed_logprob(op, value, *inputs, name=None, **kwargs):
...
-
-    trans_node = _default_transformed_rv(node.op, node)
-
-    if trans_node is None:
+    if rv_map_feature is None or values_to_transforms is None:
         return None  # pragma: no cover
 
-    # Get the old value variable and remove it from our value variables map
-    rv_var = node.outputs[1]
-    old_value_var = rv_map_feature.rv_values.pop(rv_var)
```
ricardoV94, line 118: There was no need to pop the `old_value_var` since we already have it from above and we are replacing it with the `new_value_var` at the end anyway.

aeppl/transforms.py
```diff
@@ -237,11 +299,27 @@ def log_jac_det(self, value, *inputs):
         return at.zeros(value.shape)
 
 
-def create_default_transformed_rv_op(
+def _create_transformed_rv_op(
```
ricardoV94, line 302: This method is going to be further expanded in #26 so it seems reasonable to make it private and just expose the user facing `create_default_transformed_rv_op` with the fixed default kwargs (via the partial)

brandonwillard: This change doesn't make the function private in any functional sense; it's just a naming convention.  If we're concerned about the things we're exposing to users, the best we can do is set `__all__`.

Regardless, this is a minor style-based refactoring that should be in a separate PR/commit.

brandonwillard (dismissed):
It looks like you (re)added some refactoring changes to this commit.

aeppl/transforms.py
brandonwillard: If we don't pop them, then we'll end up with a map that has entries for variables that are no longer in the graph, and&mdash;unless I'm missing something&mdash;we don't want/need those.

brandonwillard:
I put the most recent additions into a separate commit.

aeppl/transforms.py
```diff
@@ -89,52 +91,115 @@ def transformed_logprob(op, value, *inputs, name=None, **kwargs):
...
+        new_value_var.name = f"{value_var.name}_{transform.name}"
 
+    # Map both new and old rv to new value var
     new_rv_var = trans_node.outputs[1]
-
-    if old_value_var.name and getattr(transform, "name", None):
-        new_value_var.name = f"{old_value_var.name}_{transform.name}"
-
     rv_map_feature.rv_values[new_rv_var] = new_value_var
+    rv_map_feature.rv_values[rv_var] = new_value_var
```
brandonwillard, line 153: Do we need this association?

ricardoV94: It was there originally. I am not sure we do though

aeppl/transforms.py
ricardoV94: The original rewrite was reintroducing it in the dictionary, but associated with the new value. See the other comment.

aeppl/transforms.py
```diff
@@ -89,52 +91,115 @@ def transformed_logprob(op, value, *inputs, name=None, **kwargs):
...
+    else:
+        new_op = _create_transformed_rv_op(node.op, transform)()
+        trans_node = new_op.make_node(*node.inputs)
+        trans_node.outputs[1].name = node.outputs[1].name
 
     # We now assume that the old value variable represents the *transformed space*.
     # This means that we need to replace all instance of the old value variable
     # with "inversely/un-" transformed versions of itself.
-    new_value_var = transform.backward(old_value_var, *trans_node.inputs)
-    rv_map_feature.rv_values[rv_var] = new_value_var
```
ricardoV94, line 126: This was the line where it was being reintroduced to the mapping feature

ricardoV94:
I removed the old RV from the mapping and cleaned some changes that were creeping into the second commit. Let me know it it looks good @brandonwillard

brandonwillard (approved):

PR #47, opened 2021-08-15
Create factorized `joint_logprob`

ricardoV94:
The idea is that the core `joint_logprob` (here renamed to `factorized_joint_logprob`) would return a dictionary with `{value_var: logprob_term}` pairs, instead of adding them up as we currently do.

```python
import aesara.tensor as atr
from aeppl import factorized_joint_lopgrob

x_rv = atr.normal(name="x')
y_rv = atr.normal(x_rv, name="y")

factorized_joint_logprob(y_rv, {x_rv: x, y_rv: y})
```
```python
# returns
{
    y: normal_logprob(y, x, 1),
    x: normal_logprob(x, 0, 1),
}
```
CC @brandonwillard

### DIFF
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -207,11 +207,18 @@ def factorized_joint_logprob(
     return logprob_vars
 
 
-def joint_logprob(*args, **kwargs) -> Optional[TensorVariable]:
+def joint_logprob(*args, sum: bool = True, **kwargs) -> Optional[TensorVariable]:
     """Create a graph representing the joint log-probability/measure of a graph.
 
     This function calls `factorized_joint_logprob` and returns the combined
     log-probability factors as a single graph.
+
+    Parameters
+    ----------
+    sum: bool
+        If ``True`` each factor is collapsed to a scalar via ``sum`` before
+        being joined with the remaining factors. This may be necessary to
+        avoid incorrect broadcasting among independent factors.
 
     """
     logprob = factorized_joint_logprob(*args, **kwargs)
@@ -219,6 +226,12 @@ def joint_logprob(*args, **kwargs) -> Optional[TensorVariable]:
         return None
     elif len(logprob) == 1:
         logprob = tuple(logprob.values())[0]
-        return at.sum(logprob)
+        if sum:
+            return at.sum(logprob)
+        else:
+            return logprob
     else:
-        return at.add(*[at.sum(factor) for factor in logprob.values()])
+        if sum:
+            return at.sum([at.sum(factor) for factor in logprob.values()])
+        else:
+            return at.add(*logprob.values())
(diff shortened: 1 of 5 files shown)


aeppl/joint_logprob.py
```diff
@@ -173,20 +175,35 @@ def joint_logprob(
...
+
+                if q_rv_value_var not in logprob_vars:
+                    logprob_vars[q_rv_value_var] = q_logprob_var
                 else:
-                    logprob_var += q_logprob_var
+                    warnings.warn(
+                        f"The value var {q_rv_value_var} was assigned more than one logprob factor",
+                        RuntimeWarning,
+                    )
+                    logprob_vars[q_rv_value_var] += q_logprob_var
```
ricardoV94, line 193: Or append to a list?

brandonwillard: I don't think it should.

aeppl/joint_logprob.py
```diff
@@ -17,7 +18,7 @@
 from aeppl.utils import rvs_to_value_vars
 
 
-def joint_logprob(
+def factorized_joint_logprob(
```
ricardoV94, line 21: It's possible we could do away with the `vars` argument entirely, since these are available as the keys of `rv_values`

tests/test_transforms.py
```diff
@@ -188,7 +188,7 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
             )
 
         exp_logprob_val += np.log(np.linalg.det(jacobian_val))
-        exp_logprob_val += b_dist.logpdf(b_val)
+        exp_logprob_val += b_dist.logpdf(b_val).sum()
```
ricardoV94, line 191: Here we had an incorrect logprob in the `dirichlet` test, because the normal `b_dist (size=2)` was broadcasting the otherwise scalar `a_dist.logpdf(a_val) (size=())`

aeppl/joint_logprob.py
```diff
@@ -201,16 +201,26 @@ def factorized_joint_logprob(
...
     """Create a graph representing the joint log-probability/measure of a graph.
 
     This function calls `factorized_joint_logprob` and returns the combined
     log-probability factors as a single graph.
 
+    :param sum:
+        If ``True`` each factor is collapsed to a scalar via ``sum`` before
+        being joined with the remaining factors. This may be necessary to
+        avoid incorrect broadcasting among independent factors.
+
```
brandonwillard, line 214: We've been using the NumPy docstring format.

aeppl/joint_logprob.py
brandonwillard: That sounds correct.

brandonwillard (changes requested):
Just some minor changes; otherwise, this looks good.

https://github.com/aesara-devs/aeppl/pull/53 should fix the sum issue alone, so you might not need the test changes in the first commit after that's merged.

ricardoV94:
Shall I keep the sum optional?

ricardoV94:
I reverted the changes added in #53 to the tests now that `sum` is optional.

PR #54, opened 2021-08-23
Fix issues involving "indirect" random variables

brandonwillard:
This PR addresses issues with value variable tracking for random variables that aren't passed directly to `joint_logprob`.


brandonwillard:
From this I noticed this Aesara issue https://github.com/aesara-devs/aesara/issues/569.

PR #58, opened 2021-09-05
Update README example

ricardoV94:
Some docstrings and README changes that were missing from #57

### DIFF
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -24,11 +24,11 @@ def factorized_joint_logprob(
     extra_rewrites: Optional[Union[GlobalOptimizer, LocalOptimizer]] = None,
     **kwargs,
 ) -> Dict[TensorVariable, TensorVariable]:
-    r"""Create a graph representing the joint log-probability/measure of a graph.
+    r"""Create a map between variables and their log-probabilities such that the
+    sum is their joint log-probability.
 
-    TODO: change this
-    The input `var` determines which graph is used and `rv_values` specifies
-    the resulting measure-space graph's input parameters.
+    The `rv_values` dictionary specifies a joint probability graph defined by
+    pairs of random variables and respective measure-space input parameters
 
     For example, consider the following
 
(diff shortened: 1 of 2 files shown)


aeppl/joint_logprob.py
```diff
@@ -26,9 +26,8 @@ def factorized_joint_logprob(
 ) -> Dict[TensorVariable, TensorVariable]:
     r"""Create a graph representing the joint log-probability/measure of a graph.
```
brandonwillard, line 27:
```suggestion
    r"""Create a map between variables and their log-probabilities such that the sum is their joint log-probability.
```

aeppl/joint_logprob.py
```diff
@@ -47,11 +46,11 @@ def factorized_joint_logprob(
         Y \sim& \operatorname{N}(0, \sigma^2)
 
     If we create a value variable for ``Y_rv``, i.e. ``y = at.scalar("y")``,
```
brandonwillard, line 48:
```suggestion
    If we create a value variable for ``Y_rv``, i.e. ``y_vv = at.scalar("y")``,
```

This only adds the `*_vv` suffix convention for value variables.

aeppl/joint_logprob.py
```diff
@@ -47,11 +46,11 @@ def factorized_joint_logprob(
...
-    the graph of ``joint_logprob({Y_rv: y})`` is equivalent to the
-    conditional probability :math:`\log p(Y = y \mid \sigma^2)`.  If we specify
-    a value variable for ``sigma2_rv``, i.e. ``s = at.scalar("s2")``, then
-    ``joint_logprob({Y_rv: y, sigma2_rv: s})`` yields the joint
-    log-probability
+    the graph of ``factorized_joint_logprob({Y_rv: y})`` is equivalent to the
+    conditional probability :math:`\log p(Y = y \mid \sigma^2)`, with a stochastic
+    ``sigma2_rv``. If we specify a value variable for ``sigma2_rv``, i.e.
+    ``s = at.scalar("s2")``, then ``factorized_joint_logprob({Y_rv: y, sigma2_rv: s})``
+    yields the joint log-probability of the two variables.
```
brandonwillard, line 53:
```suggestion
    the graph of ``factorized_joint_logprob({Y_rv: y_vv})`` is equivalent to the
    conditional probability :math:`\log p(Y = y \mid \sigma^2)`, with a stochastic
    ``sigma2_rv``. If we specify a value variable for ``sigma2_rv``, i.e.
    ``s_vv = at.scalar("s2")``, then ``factorized_joint_logprob({Y_rv: y_vv, sigma2_rv: s_vv})``
    yields the joint log-probability of the two variables.
```

This only adds the `*_vv` suffix convention for value variables.

brandonwillard (changes requested):
Since `joint_logprob` will be a likely point-of-entry for most people, we should make its signature more explicit and find a way to reuse parts of the docstring from `factorized_joint_logprob`.

Is anyone aware of a good approach to the latter that doesn't involve copy-pasting text?

ricardoV94:
> Since `joint_logprob` will be a likely point-of-entry for most people, we should make its signature more explicit and find a way to reuse parts of the docstring from `factorized_joint_logprob`.
> 
> Is anyone aware of a good approach to the latter that doesn't involve copy-pasting text?

Only thing I could find was the `functools.wraps` but doesn't really apply to this. So unless someone has any ideas, either copy-paste or place the docstrings in the `joint_logprob`

brandonwillard (approved):

PR #71, opened 2021-10-12
Restrict `BroadcastTo` lifting of `RandomVariable`s

brandonwillard:
This PR restricts the cases in which `BroadcastTo` `Op`s will be lifted through `RandomVariable` `Op`s via `naive_bcast_rv_lift`.

Simply put, an expression like `at.broadcast_to(at.random.normal(0, 1), (10,))` should not be lifted, since it would result in ten independent random variates, instead of a single variate that's broadcasted to the shape `(10,)`.  Lifting should only happen when the shape of the `RandomVariable` already matches the broadcasted shape (e.g. via `size` and/or one of the distribution parameters), or when the broadcasting only introduces broadcastable dimensions (i.e. adds extra dimensions of length one).

This PR is&mdash;in part&mdash;an answer to an issue mentioned in Part 2 of this comment: https://github.com/aesara-devs/aeppl/discussions/51#discussioncomment-1192490.

Currently, this draft PR only introduces some tests for `naive_bcast_rv_lift` that state what needs to be changed/implemented.

- [ ] Actually put the lifting restriction in place.

### DIFF
--- a/aeppl/opt.py
+++ b/aeppl/opt.py
@@ -108,9 +108,28 @@ def incsubtensor_rv_replace(fgraph, node):
 
 @local_optimizer([BroadcastTo])
 def naive_bcast_rv_lift(fgraph, node):
-    """Lift a ``BroadcastTo`` through a ``RandomVariable`` ``Op``.
+    r"""Lift a `BroadcastTo` through a `RandomVariable` `Op`.
 
-    XXX: This implementation simply broadcasts the ``RandomVariable``'s
+    This can only be done under the assumption that the `BroadcastTo` is
+    redundant.  For example, the following are redundant `BroadcastTo`s:
+
+    .. code-block:: python
+
+        Y_1 = at.broadcast_to(at.random.normal(0, 1, size=10), (10,))
+        Y_2 = at.broadcast_to(at.random.normal(at.zeros((10,)), 1), (10,))
+        Y_3 = at.broadcast_to(at.random.normal(0, 1, size=10), (10, 1))
+
+    but the following is not:
+
+    .. code-block:: python
+
+        Y_4 = at.broadcast_to(at.random.normal(0, 1), (10,))
+
+
+    The problem with the latter is that the lifting would introduce new
+    variates.
+
+    XXX: This implementation simply broadcasts the `RandomVariable`'s
     parameters, which won't always work (e.g. multivariate distributions).
 
     TODO: Instead, it should use ``RandomVariable.ndim_supp``--and the like--to
--- a/tests/test_opt.py
+++ b/tests/test_opt.py
@@ -1,0 +1,42 @@
+import aesara.tensor as at
+import pytest
+from aesara.graph.opt import EquilibriumOptimizer
+from aesara.graph.opt_utils import optimize_graph
+from aesara.tensor.extra_ops import BroadcastTo
+
+from aeppl.opt import naive_bcast_rv_lift
+
+bcast_lift_opt = EquilibriumOptimizer(
+    [naive_bcast_rv_lift], ignore_newtrees=False, max_use_ratio=1000
+)
+
+
+@pytest.mark.parametrize(
+    "rv_params, rv_size, bcast_shape, should_rewrite",
+    [
+        # The `BroadcastTo` shouldn't be lifted, because it would imply that there
+        # are 10 independent samples, when there's really only one
+        pytest.param(
+            (0, 1),
+            None,
+            (10,),
+            False,
+            marks=pytest.mark.xfail(reason="Not implemented"),
+        ),
+        # These should work, under the assumption that `size == 10`, of course.
+        ((0, 1), at.iscalar("size"), (10,), True),
+        ((0, 1), at.iscalar("size"), (1, 10, 1), True),
+        ((at.zeros((at.iscalar("size"),)), 1), None, (10,), True),
+    ],
+)
+def test_naive_bcast_rv_lift(rv_params, rv_size, bcast_shape, should_rewrite):
+    graph = at.broadcast_to(at.random.normal(*rv_params, size=rv_size), bcast_shape)
+
+    assert isinstance(graph.owner.op, BroadcastTo)
+
+    new_graph = optimize_graph(graph, custom_opt=bcast_lift_opt)
+
+    if should_rewrite:
+        assert not isinstance(new_graph.owner.op, BroadcastTo)
+    else:
+        assert isinstance(new_graph.owner.op, BroadcastTo)


brandonwillard:
While looking into this, I realized that we can use this naive broadcasting as we always have, but that we should replace the underlying `RandomVariable`s with a duplicate `type` that doesn't allow sampling and doesn't have an associated RNG object.

The idea is that we're only using these `RandomVariable`s for log-probability derivations, so, while `at.broadcast_to(at.random.normal(0, 1), (10,))` isn't equivalent to `at.random.normal(at.broadcast_to(0, (10,)), at.broadcast_to(1, (10,)))` or `at.random.normal(0, 1, size=(10,))`, it **is** equivalent to those latter forms when the scalar value variable is also broadcasted and only a log-probability is constructed from those terms.  

In other words, we have equivalence "modulo" a log-probability generating function and an accompanying value variable transform.

To be responsible about such a rewrite, we shouldn't return `RandomVariable`s, though.  Also, by stripping the RNG objects from `RandomVariable`s we might also be able to get rid of some of those annoying and unused shared variables (e.g. the ones introduced by `Scan`).

cc: @ricardoV94 @kc611

ricardoV94:
> it is equivalent to those latter forms when the scalar value variable is also broadcasted and only a log-probability is constructed from those terms.

Broadcasting the value variable helps with not introducing multiple values where only one should exist, but we would still be double counting the same logprob term, the way aeppl works now.

Maybe we need an explicit MeasurableBroadcast that undoes the broadcast of the value variable / variable parameters before calling the base `logprob`?

brandonwillard:
> It is equivalent to the final logprob returned by aeppl, but that final logprob is not consistent with the original graph. It would double count terms.

Under our log-probability mapping, a graph representing an array containing a broadcasted random variable  would map to an array of said random variable's scalar log-probability broadcasted, so we want the result to have multiple terms (i.e be an array).

Are you refering to another issue?

Broadly, I'm talking about the implementation of a general log-probability mapping for graphs representing broadcasted random variables.  More specifically, the topic is the relevant parsing and representation of these domain elements.

ricardoV94:
> Under our log-probability mapping, a graph representing an array containing a broadcasted random variable would map to an array of said random variable's scalar log-probability broadcasted, so we want the result to have multiple terms (i.e be an array).

If I understand you correctly it means we will treat the graph `rv = at.broadcast_to(at.random.normal(0, 1), (10,))` and `at.random.normal(0, 1, size=10)` as the same?

That simplifies things quite a lot for other derived RVs, but we should decide that explicitly. 

It means we use the original aesara graph as something that somewhat *loosely* defines a log-probability graph and not as a true generative graph that we carefully invert to obtain the corresponding probability graph.

***

And then your point about the `ValuedRVs` is just to avoid distorting RVs that stay in the final graph?

brandonwillard:
> If I understand you correctly it means we will treat the two RVs as the same:

The actual equivalence is
```python
logprob(broadcast_to(Y, s), broadcast_to(y, s)) == broadcast_to(logprob(Y, y), s)
```
for `y ~ Y` (i.e. `y` is a value variable of the measurable/random variable `Y`).

I'm talking about implementing that equivalence relation above.  

In practice, we start with the expression `broadcast_to(Y, s)` and the pair `(Y, y)`, and we need to produce the left-hand side of the relation above so that we can use the right-hand side and derive the result in terms of `logprob(Y, y)`.

What I'm saying is that we can use an intermediate representation in the space of `(Y, y)`, so that `broadcast_to((Y, y), s) == (broadcast_to(Y), broadcast_to(y))` and the `loprob` operator can be defined over such pairs.  The original equivalence can be realized under this representation, as well, but it's easier to translate the existing rewrite to something that operates on `(Y, y)`, and it fits better into our parsing strategy.

My second point is about modeling `(Y, y)` pairs explicitly in the graph via an `Op`.

brandonwillard:
> I don't see what that achieves just yet. 

Aside from clarifying/understanding what it is we're actually doing and what we intend to do, using the types that are true to the system(s) they model avoids confusion, design quagmires, over-engineering, etc.  In other words, it makes it possible to continually simplify a process instead of complicating it over time (e.g. through modularity and the like).

For instance, if we sort these things out at a high level like this, we can probably find a redesign that recasts all the operations in terms of simple(r) local and global rewrites in the `[Global|Local]Optimization` framework.  Then, by making independent improvements to that framework, we automatically get improvements in our log-probability rewrites.

At a lower level, using the correct types generally implies that type-related implementation details are addressed, such as type/instance equivalence (e.g. `__eq__`) and inheritance (e.g. `isinstance`, `issubclass`), which can functionally resolve confusions and help guard against conceptual mistakes.  This is directly relevant to the topic of this PR, because, if `naive_bcast_rv_lift` operated on objects of a model-theoretically correct type, there would little room to make the mistake of applying the rewrite to `RandomVariable`s and producing inconsistent graphs.


> Seems like that already happens by the automatic broadcasting of the logprob terms when `y` is a vector of shape `s`.

Yes, it should, and that's part of the reason why a change to the basic rewrite logic in `naive_bcast_rv_lift` isn't really a solution to the issue/concerns of this PR.


> My question was about the case where `y.shape == broadcast_to(Y, s)`. In that case we might not want to broadcast the logprob term (depending on how we interpret the role of aeppl).

The premise is that we want the broadcasting, because those are the elements we're handling with the `naive_bcast_rv_lift` rewrite.  There should never be a case in which `naive_bcast_rv_lift` encounters a graph for which it can, but shouldn't, rewrite.  That would imply there's an issue with `naive_bcast_rv_lift`'s use, and not `naive_bcast_rv_lift`'s implementation.

brandonwillard:
Just to follow this chain of thought a little more, if we use the old `ObservedRV` approach (i.e. an `Op` that takes the `MeasurableVariable` and value variable as inputs), we could replace the value variable logic in `factorized_joint_logprob` with a single rewrite that replaces (_assumedly_) measurable terms with `ObservedRV`s.  

Most of our custom rewrites could be easily rewritten to handle the extra node between the graphs they rewrite; however, some of the existing `RandomVariable` rewrites (e.g. `local_dimshuffle_rv_lift`, `local_subtensor_rv_lift`) would no longer apply, again due to that extra `ObservedRV` node.

For example, the `*Subtensor*` lifting for `RandomVariable`s wouldn't work once we replaced the `RandomVariable`s with `ObservedRV`s, because any `Y_rv[idx]` graphs would be rewritten to `observed(Y_rv, y_vv)[idx]` and a rewrite would need to be created for that specific form as well.  

Fortunately, with some simple stand-alone `ObservedRV` canonicalizations, we can probably bridge that gap quite easily.  Using the previous example, we could lift `*Subtensor*` operations through `ObservedRV`s, and, after such `ObservedRV`-specific canonicalizations are applied, the more generic `RandomVariable` canonicalizations/rewrites can be applied.

We could also replace `rvs_to_value_vars` with a simple rewrite that substitutes `ObservedRV`s with their value variable inputs.

ricardoV94:
Let's give it a try and see how it evolves?

PR #72, opened 2021-10-14
Cumsum rvs

ricardoV94:
```python
import aesara.tensor as at

rv = at.random.normal(0, 1, size=5).cumsum()
vv = rv.clone()

logp = joint_logprob({rv: vv}, sum=False)
logp.eval({vv: np.arange(1, 6)})
# array([-1.41893853, -1.41893853, -1.41893853, -1.41893853, -1.41893853])
```

### DIFF
--- a/aeppl/__init__.py
+++ b/aeppl/__init__.py
@@ -11,6 +11,7 @@ from aeppl.printing import latex_pprint, pprint
 
 # isort: off
 # Add optimizations to the DBs
+import aeppl.cumsum
 import aeppl.mixture
 import aeppl.scan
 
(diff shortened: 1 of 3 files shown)


ricardoV94:
> Looks good, but I don't think the CumSum code should be in scan.py—that module is pretty specific to Scan.

Yeah, I was not very happy with it tucked in there either. Aesara puts this is `extra_ops` though, should we do the same or something more direct?

brandonwillard:
> Yeah, I was not very happy with it tucked in there either. Aesara puts this is `extra_ops` though, should we do the same or something more direct?

An `extra_ops` module is fine for now.

> Edit: Otherwise, we can organize by conceptual types of RVs... like the mixtures. 

Exactly

brandonwillard (approved):
Looks great!

PR #78, opened 2021-10-23
Introduce `ValuedVariable`

brandonwillard:
This PR replaces #71 and addresses the issues and discussions therein.

The `ValuedVariable` `Op` adds the value variable to the graph so that `PreserveRVMappings` is no longer needed.  It also clarifies the definition and actions of rewrites that truly apply to a `MeasurableVariable` and its value variable simultaneously by making `ValuedVariable`s the targets and outputs of such rewrites.

For instance, the old `naive_bcast_rv_lift` can be refactored into a rewrite that takes and produces `ValuedVariable`s, and the `MeasurableVariable`/`RandomVariable` terms being broadcasted within the `ValuedVariable` can be lifted *and* have their corresponding scalar value variables broadcasted all in the same place and way.  This approach clarifies the nature of the rewrite and makes the entire operation consistent with respect to the relation that it represents.

Our current approach (i.e. `naive_bcast_rv_lift`) is inconsistent as a stand-alone rewrite, because it doesn't also handle the value variable broadcasting.  Instead, that is handled much further down the line within `_logprob` implementations; however, that approach is fraught with potentially hard(er)-to-track issues.

- [x] Refactor user-specified transforms (i.e. `TransformValuesRewrite`)
    Our whole approach to this needs to be rewritten, mostly because the current one is too convoluted.  We can simply have a helper function that performs a `FunctionGraph.replace_all` and adds Jacobian terms to the appropriate variables directly.
- [x] Finish refactoring `*Subtensor*` support
- [x] Finish refactoring `Scan` support
- [ ] Change use of the term "valued" to "bound"
- [x] Determine how we want to handle `test_persist_inputs`, `test_warn_random_not_found`, `test_multiple_rvs_to_same_value_raises`, and `test_fail_multiple_clip_single_base`

### DIFF
--- a/aeppl/rewriting.py
+++ b/aeppl/rewriting.py
@@ -2,7 +2,7 @@ from typing import Dict, Optional, Tuple
 
 import aesara.tensor as at
 from aesara.compile.mode import optdb
-from aesara.graph.basic import Variable
+from aesara.graph.basic import Apply, Variable
 from aesara.graph.features import Feature
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.rewriting.basic import GraphRewriter, node_rewriter
@@ -20,9 +20,8 @@ from aesara.tensor.subtensor import (
     IncSubtensor,
     Subtensor,
 )
-from aesara.tensor.var import TensorVariable
 
-from aeppl.abstract import MeasurableVariable
+from aeppl.abstract import MeasurableVariable, ValuedVariable, valued_variable
 from aeppl.dists import DiracDelta
 from aeppl.utils import indices_from_subtensor
 
@@ -44,37 +43,15 @@ class NoCallbackEquilibriumDB(EquilibriumDB):
         return res
 
 
-class PreserveRVMappings(Feature):
-    r"""Keeps track of random variables and their respective value variables during
-    graph rewrites in `rv_values`
+class MeasurableConversionTracker(Feature):
+    r"""Keeps track of variables that were converted to `MeasurableVariable`\s.
 
-    When a random variable is replaced in a rewrite, this `Feature` automatically
-    updates the `rv_values` mapping, so that the new variable is linked to the
-    original value variable.
-
-    In addition this `Feature` provides functionality to manually update a random
-    and/or value variable. A mapping from the transformed value variables to the
-    the original value variables is kept in `original_values`.
-
-    Likewise, a `measurable_conversions` map is maintained, which holds
+    A `measurable_conversions` map is tracked that holds
     information about un-valued and un-measurable variables that were replaced
     with measurable variables.  This information can be used to revert these
     rewrites.
 
     """
-
-    def __init__(self, rv_values: Dict[TensorVariable, TensorVariable]):
-        """
-        Parameters
-        ==========
-        rv_values
-            Mappings between random variables and their value variables.
-            The keys of this map are what this `Feature` keeps updated.
-            The ``dict`` is updated in-place.
-        """
-        self.rv_values = rv_values
-        self.original_values = {v: v for v in rv_values.values()}
-        self.measurable_conversions: Dict[Variable, Variable] = {}
 
     def on_attach(self, fgraph):
         if hasattr(fgraph, "preserve_rv_mappings"):
@@ -82,55 +59,21 @@ class PreserveRVMappings(Feature):
                 f"{fgraph} already has the `PreserveRVMappings` feature attached."
             )
 
-        fgraph.preserve_rv_mappings = self
-
-    def update_rv_maps(
-        self,
-        old_rv: TensorVariable,
-        new_value: TensorVariable,
-        new_rv: Optional[TensorVariable] = None,
-    ):
-        """Update mappings for a random variable.
-
-        It also creates/updates a map from new value variables to their
-        original value variables.
-
-        Parameters
-        ==========
-        old_rv
-            The random variable whose mappings will be updated.
-        new_value
-            The new value variable that will replace the current one assigned
-            to `old_rv`.
-        new_rv
-            When non-``None``, `old_rv` will also be replaced with `new_rv` in
-            the mappings, as well.
-        """
-        old_value = self.rv_values.pop(old_rv)
-        original_value = self.original_values.pop(old_value)
-
-        if new_rv is None:
-            new_rv = old_rv
-
-        self.rv_values[new_rv] = new_value
-        self.original_values[new_value] = original_value
+        fgraph.measurable_conversions = {}
 
     def on_change_input(self, fgraph, node, i, r, new_r, reason=None):
-        """
-        Whenever a node is replaced during rewrite, we check if it had a value
-        variable associated with it and map it to the new node.
-        """
-        r_value_var = self.rv_values.pop(r, None)
-        if r_value_var is not None:
-            self.rv_values[new_r] = r_value_var
-        elif (
-            new_r not in self.rv_values
-            and r.owner
+        if (
+            r.owner
             and new_r.owner
-            and not isinstance(r.owner.op, MeasurableVariable)
+            and not isinstance(r.owner.op, (ValuedVariable, MeasurableVariable))
             and isinstance(new_r.owner.op, MeasurableVariable)
+            and not any(
+                isinstance(node.op, ValuedVariable)
+                for node, idx in fgraph.clients[new_r]
+                if isinstance(node, Apply)
+            )
         ):
-            self.measurable_conversions[r] = new_r
+            fgraph.measurable_conversions[r] = new_r
 
 
 @register_canonicalize
@@ -165,7 +108,7 @@ def local_remove_DiracDelta(fgraph, node):
     return [dd_val]
 
 
-@node_rewriter(inc_subtensor_ops)
+@node_rewriter([ValuedVariable])
 def incsubtensor_rv_replace(fgraph, node):
     r"""Replace `*IncSubtensor*` `Op`\s and their value variables for log-probability calculations.
 
@@ -183,37 +126,32 @@ def incsubtensor_rv_replace(fgraph, node):
 
     This provides a means of specifying "missing data", for instance.
     """
-    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+    incsubtensor_var, value_var = node.inputs
+    subtensor_node = incsubtensor_var.owner
 
-    if rv_map_feature is None:
+    if subtensor_node is None or not isinstance(subtensor_node.op, inc_subtensor_ops):
         return None  # pragma: no cover
 
-    if not isinstance(node.op, inc_subtensor_ops):
-        return None  # pragma: no cover
-
-    rv_var = node.outputs[0]
-    if rv_var not in rv_map_feature.rv_values:
-        return None  # pragma: no cover
-
-    base_rv_var = node.inputs[0]
+    base_rv_var = subtensor_node.inputs[0]
 
     if not (
         base_rv_var.owner
         and isinstance(base_rv_var.owner.op, MeasurableVariable)
-        and base_rv_var not in rv_map_feature.rv_values
+        and not isinstance(base_rv_var, ValuedVariable)
     ):
         return None  # pragma: no cover
 
-    data = node.inputs[1]
-    idx = indices_from_subtensor(getattr(node.op, "idx_list", None), node.inputs[2:])
+    data = subtensor_node.inputs[1]
+    idx = indices_from_subtensor(
+        getattr(subtensor_node.op, "idx_list", None), subtensor_node.inputs[2:]
+    )
 
     # Create a new value variable with the indices `idx` set to `data`
-    value_var = rv_map_feature.rv_values[rv_var]
     new_value_var = at.set_subtensor(value_var[idx], data)
-    rv_map_feature.update_rv_maps(rv_var, new_value_var, base_rv_var)
 
-    # Return the `RandomVariable` being indexed
-    return [base_rv_var]
+    new_base_rv_var = valued_variable(base_rv_var, new_value_var)
+
+    return [new_base_rv_var]
 
 
 logprob_rewrites_db = SequenceDB()
@@ -248,7 +186,7 @@ def construct_ir_fgraph(
     r"""Construct a `FunctionGraph` in measurable IR form for the keys in `rv_values`.
 
     A custom IR rewriter can be specified. By default,
-    `logprob_rewrites_db.query(RewriteDatabaseQuery(include=["basic"]))` is used.
+    ``logprob_rewrites_db.query(RewriteDatabaseQuery(include=["basic"]))`` is used.
 
     Our measurable IR takes the form of an Aesara graph that is more-or-less
     equivalent to a given Aesara graph (i.e. the keys of `rv_values`) but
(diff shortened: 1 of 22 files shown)


aeppl/rewriting.py
```diff
@@ -282,44 +222,78 @@ def construct_ir_fgraph(
...
+    # TODO: It's not clear that this is actually needed.
+    rv_vars = list(rv_value_clones.keys())
+    # sorted_rv_vars: List[TensorVariable] = sum(
+    #     [
+    #         [o for o in node.outputs if o in rv_vars]
+    #         for node in io_toposort((), rv_vars)
+    #     ],
+    #     [],
+    # )
+    sorted_rv_vars = rv_vars
```
rlouf, line 255: Is this still necessary?

rlouf: If anything we should perform operations on variables that are conditioned on but are not `RandomVariables` first. Their value variable should be used to set the expression of the value variable of the `RandomVariable` upstream so we can support the examples in https://github.com/aesara-devs/aeppl/discussions/119

rlouf:
```python
import aesara.tensor as at

x = at.normal()
y = at.log(x)
z = at.normal(y)

y_vv = y.clone()
z_vv = z.clone()
joint_logprob({y: y_vv, z: z_vv})
```

In this example we should also replace `x` with a `ValueVariable` whose value variable is set to `x_vv = at.exp(y_vv)`, for instance.

aeppl/tensor.py
```diff
@@ -10,50 +10,50 @@
...
+from aeppl.rewriting import measurable_ir_rewrites_db
 
 
 @node_rewriter([BroadcastTo])
-def naive_bcast_rv_lift(fgraph, node):
-    """Lift a ``BroadcastTo`` through a ``RandomVariable`` ``Op``.
-
-    XXX: This implementation simply broadcasts the ``RandomVariable``'s
-    parameters, which won't always work (e.g. multivariate distributions).
+def valued_var_bcast_lift(fgraph, node):
```
rlouf, line 23: Naive question: why does this only return a `RandomVariable`, and not a new `ValuedVariable`? Also, why are we not setting the shape of the associated value variables as well?

tests/test_censoring.py
```diff
@@ -131,7 +132,10 @@ def test_fail_multiple_clip_single_base():
     cens_rv2 = at.clip(base_rv, -1, 1)
     cens_rv2.name = "cens2"
 
-    with pytest.raises(RuntimeError, match=r"could not be derived: {cens\d}"):
```
rlouf, line 134: This is not different from the following:

```python
base_rv = at.random.normal(0, 1)
exp_rv1 = at.exp(base_rv)
exp_rv2 = at.exp(base_rv)

conditional_logprob(exp_rv1, exp_rv2)
```

This code implies that the value of `base_rv` to two different values at the same time, which is impossible.

tests/test_joint_logprob.py
```diff
@@ -241,6 +243,7 @@ def test_warn_random_not_found():
         conditional_logprob(y_rv, warn_missing_rvs=False)
 
 
+@pytest.mark.skip(reason="Disabled until we decide how we want this to work.")
```
rlouf, line 246: I don't see why this one shouldn't work. There is no ambiguity there, the two logdensities can be bound to the same value.

tests/test_joint_logprob.py
```diff
@@ -229,6 +230,7 @@ def test_persist_inputs():
     assert y_vv in ancestors([logp_2])
 
 
+@pytest.mark.skip(reason="Disabled until we decide how we want this to work.")
```
rlouf, line 233: Wouldn't we want the logdensity to be bound to `x_rv`, i.e. it will use a different realization of `x` at each call?

tests/test_joint_logprob.py
```diff
@@ -211,6 +211,7 @@ def test_joint_logprob_subtensor():
         np.testing.assert_almost_equal(logp_vals, exp_obs_logps, decimal=decimals)
 
 
+@pytest.mark.skip(reason="Disabled until we decide how we want this to work.")
```
rlouf, line 214: What is not clear here?

aeppl/tensor.py
brandonwillard: I don't think we even need that rewrite anymore, so I'll remove it.

tests/test_censoring.py
brandonwillard: I've added an error for when multiple values are bound to the same measurable term.

tests/test_joint_logprob.py
brandonwillard: Yeah, this is an elective warning, because we definitely need to support the derivation of log-densities with un-bound/valued `RandomVariable` terms.

tests/test_joint_logprob.py
brandonwillard: It's not clear that we should guarantee the (object) identities of the value graphs bound to a measurable term.  It definitely seems like a convenient thing to have in some situations, but it shouldn't be necessary for any functionality.

tests/test_joint_logprob.py
rlouf: If this is something we want to support we should remove this warning, and instead document the behavior in the docstring (it's already the case).

tests/test_joint_logprob.py
rlouf: Thoughts?

tests/test_joint_logprob.py
rlouf: I think it would be confusing to not guarantee this identity for values that are passed as realizations of a measurable term, as is the case here. What if we observe not `Y_rv` directly, but another quantity from which its value can be derived like in the following example?

```python
import aesara
import aesara.tensor as at


x = at.scalar("x")
beta_rv = at.random.normal(0, 1, name="beta")
Y_rv = at.random.normal(beta_rv * x, 1, name="y")

obs_at = Y_rv.clone()
y_vv = at.log(obs_at + 1.)
logp, (beta_vv,) = joint_logprob(beta_rv, realized={Y_rv: y_vv})
```

Shouldn't we be able to compile this function:

```python
logp_fn = aesara.function([obs_at, beta_vv], logp)
```

brandonwillard: > Shouldn't we be able to compile this function?

Yes, and for that example, I think all that's missing is `x` as an input.

tests/test_joint_logprob.py
brandonwillard: Yeah, I agree; I'll change/update this one.

tests/test_joint_logprob.py
brandonwillard: The main issue/question behind this one is that realized value-variable graphs like `y * 2` can have canonicalization applied to them, so that the log-probability output graph(s) will not contain the original `y * 2` subgraphs, but canonicalized ones like `2 * y`.  

This isn't actually an issue or limitation of any sort, especially since a user-held "original" value-variable graph can/will be merged with its canonicalized form when both graphs are put together at some point during compilation.  I recall using the old identity guarantee a long time ago for something, but that should no longer be necessary (and it never really should have been).

Anyway, as long as the identities of **inputs** to such graphs are preserved (i.e. `y` in our example), a user should be able to compile everything as expected.

I've updated the test to reflect these sentiments.

rlouf:
Will the following snippet output the correct logdensity?

```python
import aeppl
import aesara.tensor as at

x = srng.normal(0, 1)
z = srng.normal(0, 1)

x_tr = at.log(x)
logprob, vvs  = aeppl.joint_logprob(x_tr, z)
```

AePPL currently returns a logdensity graph where the input to `z`'s conditional logdensity is a stochastic variable, instead of the exponential of the value variable bound to `x_tr`. 

If not we can tackle this in another PR.

brandonwillard:
> Will the following snippet output the correct logdensity?
> 
> ```python
> import aeppl
> import aesara.tensor as at
> 
> srng = at.random.RandomStream(0)
> 
> x = srng.normal(0, 1)
> z = srng.normal(0, 1)
> 
> x_tr = at.log(x)
> logprob, vvs  = aeppl.joint_logprob(x_tr, z)
> ```
> 
> AePPL currently returns a logdensity graph where the input to `z`'s conditional logdensity is a stochastic variable, instead of the exponential of the value variable bound to `x_tr`.
> 
> If not we can tackle this in another PR.

The following passes:
```python

Z_rv = srng.normal(0, 1, name="Z")

x_tr = at.exp(X_rv)
logprob, vvs = joint_logprob(x_tr, Z_rv)

logp_fn = aesara.function(vvs, logprob)

x_val, z_val = 0.1, 0.1
exp_res = sp.stats.lognorm(s=1).logpdf(x_val) + sp.stats.norm().logpdf(z_val)

np.testing.assert_allclose(logp_fn(x_val, z_val), exp_res)
```

brandonwillard:
This case doesn't work, though:
```python

Z_tr = at.exp(X_rv)

z_vv = at.dscalar(name="z_vv")

logprob, vvs = joint_logprob(realized={Z_tr: z_vv, X_rv: z_vv})
```

The IR for this case is as follows:
```python
ValuedVariable [id A] 1
 |Elemwise{exp,no_inplace} [id B] 0
 | |z_vv [id C]
 |z_vv [id C]
ValuedVariable [id D] 3
 |normal_rv{0, (0, 0), floatX, False}.1 [id E] 'X' 2
 | |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F2C85101660>) [id F]
 | |TensorConstant{[]} [id G]
 | |TensorConstant{11} [id H]
 | |TensorConstant{0} [id I]
 | |TensorConstant{1} [id J]
 |z_vv [id C]
```

A log-probability isn't produced because we replaced the original `X_rv` with its value variable in the graph of `Z_tr`.  We could replace the `z_vv` in the `Elemwise` with the `ValueVariable` for `X_rv` (i.e. `id D`).  That was actually the original intention of this approach, but I haven't made that addition/change yet.

brandonwillard:
> A log-probability isn't produced because we replaced the original `X_rv` with its value variable in the graph of `Z_tr`. We could replace the `z_vv` in the `Elemwise` with the `ValueVariable` for `X_rv` (i.e. `id D`). That was actually the original intention of this approach, but I haven't made that addition/change yet. We can work on that in a follow-up PR, though.

I made some small changes, and it looks like this might not be particularly difficult to add.  The value variable transform approach needs to be updated, though.

brandonwillard:
OK, all the functionality is there: we can get the conditional densities for transformed variables *and* the densities they transform.

rlouf:
Great! How about this one?

```python
import aeppl
import aesara.tensor as at

x = srng.normal(0, 1)
z = srng.normal(x, 1)

x_tr = at.log(x)
logprob, vvs  = aeppl.joint_logprob(x_tr, z)
```

(Important if we want to refactor the approach to transforms).

We should merge this even if this example doesn't pass, and do the necessary changes in another PR. This PR is already a huge improvement.

brandonwillard:
> Great! How about this one?
> 
> ```python
> import aeppl
> import aesara.tensor as at
> 
> srng = at.random.RandomStream(0)
> 
> x = srng.normal(0, 1)
> z = srng.normal(x, 1)
> 
> x_tr = at.exp(x)
> logprob, vvs  = aeppl.joint_logprob(x_tr, z)
> ```
> 
> (Important if we want to refactor the approach to transforms).
> 
> We should merge this even if this example doesn't pass, and do the necessary changes in another PR. This PR is already a huge improvement.

That example appears to work as expected.  In other words, the following succeeds:
```python
srng = at.random.RandomStream(203920)

X = srng.normal(0, 1, name="X")
Y = srng.normal(X, 1, name="Y")
Z = at.exp(X)

logprob, (z_vv, y_vv) = joint_logprob(Z, Y)

aesara.dprint(logprob)
# Sum{acc_dtype=float64} [id A]
#  |MakeVector{dtype='float64'} [id B]
#    |Sum{acc_dtype=float64} [id C]
#    | |Elemwise{add,no_inplace} [id D]
#    |   |Check{sigma > 0} [id E]
#    |   | |Elemwise{sub,no_inplace} [id F]
#    |   | | |Elemwise{sub,no_inplace} [id G]
#    |   | | | |Elemwise{mul,no_inplace} [id H]
#    |   | | | | |TensorConstant{-0.5} [id I]
#    |   | | | | |Elemwise{pow,no_inplace} [id J]
#    |   | | | |   |Elemwise{true_div,no_inplace} [id K]
#    |   | | | |   | |Elemwise{sub,no_inplace} [id L]
#    |   | | | |   | | |Elemwise{log,no_inplace} [id M]
#    |   | | | |   | | | |<TensorType(float64, ())> [id N]
#    |   | | | |   | | |TensorConstant{0} [id O]
#    |   | | | |   | |TensorConstant{1} [id P]
#    |   | | | |   |TensorConstant{2} [id Q]
#    |   | | | |Elemwise{log,no_inplace} [id R]
#    |   | | |   |Elemwise{sqrt,no_inplace} [id S]
#    |   | | |     |TensorConstant{6.283185307179586} [id T]
#    |   | | |Elemwise{log,no_inplace} [id U]
#    |   | |   |TensorConstant{1} [id P]
#    |   | |All [id V]
#    |   |   |Elemwise{gt,no_inplace} [id W]
#    |   |     |TensorConstant{1} [id P]
#    |   |     |TensorConstant{0.0} [id X]
#    |   |Elemwise{neg,no_inplace} [id Y]
#    |     |Elemwise{log,no_inplace} [id Z]
#    |       |<TensorType(float64, ())> [id N]
#    |Sum{acc_dtype=float64} [id BA]
#      |Check{sigma > 0} [id BB] 'Y_logprob'
#        |Elemwise{sub,no_inplace} [id BC]
#        | |Elemwise{sub,no_inplace} [id BD]
#        | | |Elemwise{mul,no_inplace} [id BE]
#        | | | |TensorConstant{-0.5} [id BF]
#        | | | |Elemwise{pow,no_inplace} [id BG]
#        | | |   |Elemwise{true_div,no_inplace} [id BH]
#        | | |   | |Elemwise{sub,no_inplace} [id BI]
#        | | |   | | |Y_vv [id BJ]
#        | | |   | | |normal_rv{0, (0, 0), floatX, False}.1 [id BK] 'X'
#        | | |   | |   |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FCA2701FD60>) [id BL]
#        | | |   | |   |TensorConstant{[]} [id BM]
#        | | |   | |   |TensorConstant{11} [id BN]
#        | | |   | |   |TensorConstant{0} [id O]
#        | | |   | |   |TensorConstant{1} [id P]
#        | | |   | |TensorConstant{1} [id P]
#        | | |   |TensorConstant{2} [id BO]
#        | | |Elemwise{log,no_inplace} [id BP]
#        | |   |Elemwise{sqrt,no_inplace} [id BQ]
#        | |     |TensorConstant{6.283185307179586} [id BR]
#        | |Elemwise{log,no_inplace} [id BS]
#        |   |TensorConstant{1} [id P]
#        |All [id BT]
#          |Elemwise{gt,no_inplace} [id BU]
#            |TensorConstant{1} [id P]
#            |TensorConstant{0.0} [id BV]

logp_fn = aesara.function((z_vv, y_vv), logprob)

z_val, y_val = 0.1, 0.3

mean_val = at.random.RandomStream(203920).normal().eval()
exp_res = sp.stats.lognorm(s=1).logpdf(z_val) + sp.stats.norm(loc=mean_val).logpdf(y_val)

np.testing.assert_allclose(logp_fn(z_val, y_val), exp_res)
```

rlouf:
> #        | | |   | |Elemwise{sub,no_inplace} [id BI]
> 
> #        | | |   | | |Y_vv [id BJ]
> 
> #        | | |   | | |normal_rv{0, (0, 0), floatX, False}.1 [id BK] 'X'
> 
> #        | | |   | |   |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FCA2701FD60>) [id BL]
> 
> #        | | |   | |   |TensorConstant{[]} [id BM]
> 
> #        | | |   | |   |TensorConstant{11} [id BN]
> 
> #        | | |   | |   |TensorConstant{0} [id O]
> 
> #        | | |   | |   |TensorConstant{1} [id P]
> 
> #        | | |   | |TensorConstant{1} [id P]
> 
> #        | | |   |TensorConstant{2} [id BO]
> 


Shouldn't we have `x_vv = at.exp(y_vv)` here instead of `X`?

brandonwillard:
> ```
> # Elemwise{sub,no_inplace} [id BI]
> #        | | |   | | |Y_vv [id BJ]
> #        | | |   | | |normal_rv{0, (0, 0), floatX, False}.1 [id BK] 'X'
> #        | | |   | |   |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FCA2701FD60>) [id BL]
> #        | | |   | |   |TensorConstant{[]} [id BM]
> #        | | |   | |   |TensorConstant{11} [id BN]
> #        | | |   | |   |TensorConstant{0} [id O]
> #        | | |   | |   |TensorConstant{1} [id P]
> #        | | |   | |TensorConstant{1} [id P]
> #        | | |   |TensorConstant{2} [id BO]
> ```
> 
> Shouldn't we have `X_vv = at.exp(Z_vv)` here instead of `X`?

No, I don't think so.  As I understand it, `Z = at.exp(X)` is the only transformed and valued/bound term for which we're requesting a conditional density in that example, and the other conditional density (i.e. for `Y`) shouldn't be affected.

You might be asking about "value/bound variable transforms" (i.e. the transforms constructed by `TransformValuesRewrite` that change all instances of their transformed values).  We really need a good, distinct name for those transforms, in contrast to the kinds of "transforms" that simply produce a density for a measurable function of measurable variables.

brandonwillard:
Here's an example of `TransformValuesRewrite` that might be what you're looking for:
```python
srng = at.random.RandomStream(203920)

X = srng.normal(0, 1, name="X")
Y = srng.normal(X, 1, name="Y")

transform_rewrite = TransformValuesRewrite({X: ExpTransform()})

logprob, (x_vv, y_vv) = joint_logprob(X, Y, extra_rewrites=transform_rewrite)

aesara.dprint(logprob)
# Sum{acc_dtype=float64} [id A]
#  |MakeVector{dtype='float64'} [id B]
#    |Sum{acc_dtype=float64} [id C]
#    | |Elemwise{add,no_inplace} [id D] 'X_logprob'
#    |   |Check{sigma > 0} [id E]
#    |   | |Elemwise{sub,no_inplace} [id F]
#    |   | | |Elemwise{sub,no_inplace} [id G]
#    |   | | | |Elemwise{mul,no_inplace} [id H]
#    |   | | | | |TensorConstant{-0.5} [id I]
#    |   | | | | |Elemwise{pow,no_inplace} [id J]
#    |   | | | |   |Elemwise{true_div,no_inplace} [id K]
#    |   | | | |   | |Elemwise{sub,no_inplace} [id L]
#    |   | | | |   | | |Elemwise{log,no_inplace} [id M]
#    |   | | | |   | | | |X_vv-trans [id N]
#    |   | | | |   | | |TensorConstant{0} [id O]
#    |   | | | |   | |TensorConstant{1} [id P]
#    |   | | | |   |TensorConstant{2} [id Q]
#    |   | | | |Elemwise{log,no_inplace} [id R]
#    |   | | |   |Elemwise{sqrt,no_inplace} [id S]
#    |   | | |     |TensorConstant{6.283185307179586} [id T]
#    |   | | |Elemwise{log,no_inplace} [id U]
#    |   | |   |TensorConstant{1} [id P]
#    |   | |All [id V]
#    |   |   |Elemwise{gt,no_inplace} [id W]
#    |   |     |TensorConstant{1} [id P]
#    |   |     |TensorConstant{0.0} [id X]
#    |   |Elemwise{neg,no_inplace} [id Y]
#    |     |Elemwise{log,no_inplace} [id Z]
#    |       |X_vv-trans [id N]
#    |Sum{acc_dtype=float64} [id BA]
#      |Check{sigma > 0} [id BB] 'Y_logprob'
#        |Elemwise{sub,no_inplace} [id BC]
#        | |Elemwise{sub,no_inplace} [id BD]
#        | | |Elemwise{mul,no_inplace} [id BE]
#        | | | |TensorConstant{-0.5} [id BF]
#        | | | |Elemwise{pow,no_inplace} [id BG]
#        | | |   |Elemwise{true_div,no_inplace} [id BH]
#        | | |   | |Elemwise{sub,no_inplace} [id BI]
#        | | |   | | |Y_vv [id BJ]
#        | | |   | | |TransformedVariable [id BK]
#        | | |   | |   |Elemwise{log,no_inplace} [id M]
#        | | |   | |   |X_vv-trans [id N]
#        | | |   | |TensorConstant{1} [id P]
#        | | |   |TensorConstant{2} [id BL]
#        | | |Elemwise{log,no_inplace} [id BM]
#        | |   |Elemwise{sqrt,no_inplace} [id BN]
#        | |     |TensorConstant{6.283185307179586} [id BO]
#        | |Elemwise{log,no_inplace} [id BP]
#        |   |TensorConstant{1} [id P]
#        |All [id BQ]
#          |Elemwise{gt,no_inplace} [id BR]
#            |TensorConstant{1} [id P]
#            |TensorConstant{0.0} [id BS]

logp_fn = aesara.function((x_vv, y_vv), logprob)

x_val, y_val = np.exp(0.1), 0.3

exp_res = sp.stats.lognorm(s=1).logpdf(x_val) + sp.stats.norm(loc=np.log(x_val)).logpdf(y_val)

np.testing.assert_allclose(logp_fn(x_val, y_val), exp_res)
```

rlouf:
My ideas aren't completely clear, although I still feel what I described should be the expected behavior. Anyway, this PR is not the best place to discuss this. I suggest we merge this PR and open a separate issue/discussion.

brandonwillard:
OK, I'm going to merge this now and follow up with a few more improvements shortly.

PR #82, opened 2021-10-29
Allow both edges of IntervalTransform to be None

ricardoV94:
This situation cropped up in https://github.com/pymc-devs/pymc/pull/5087

In general, it seems useful to allow for the possibility of a useless IntervalTransform when working with truncated distributions

### DIFF
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -224,6 +224,8 @@ class IntervalTransform(RVTransform):
             return at.log(value - a)
         elif b is not None:
             return at.log(b - value)
+        else:
+            return value
 
     def backward(self, value, *inputs):
         a, b = self.args_fn(*inputs)
@@ -235,6 +237,8 @@ class IntervalTransform(RVTransform):
             return at.exp(value) + a
         elif b is not None:
             return b - at.exp(value)
+        else:
+            return value
 
     def log_jac_det(self, value, *inputs):
         a, b = self.args_fn(*inputs)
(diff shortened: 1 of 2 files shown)


brandonwillard:
This does not appear to be a reasonable approach to the situation.  In general, useless `Op`s/transforms need to be removed from a graph entirely and not intentionally added; otherwise, we add complexity to the graph representation and all the downstream logic that deals with such inconsistencies (e.g. *all* Aesara graph traversal and rewriting that needs to deal with these no-op `Op`s/sub-graphs/code).

In other words, the result of these changes may have been a trade between a potentially minor PyMC-specific use-case inconvenience for an inherent canonicalization issue in Aesara/AePPL.

If there's important/necessary information being encoded/conveyed by such no-op transforms, it should be detailed in the PR description and/or commits.  In the meantime, these commits will need to be reverted until that information is forthcoming and a case is made for the no-ops.

ricardoV94:
Sure we can revert it, but in practice this was just completing the 4 logical possibilities of None x None, which we were implicitly using to encode -inf, +inf bounds to define one sided transforms.

Without it the IntervalTransform logic would just fall off and return None for the forward and backward methods, while returning value for the log_jac_det.

It feels more internally consistent. The fact that it happens to be useless seems less important.

brandonwillard:
Let's start with the details of the situation and see if that's necessary before we revert&mdash;that's why I made #82 instead of reverting (and so we don't forget to follow up on this).

There's a point to be made about runtime values of symbolic variables making the intervals zero, but, beyond that, we need more justification (e.g. are these results consistent when the endpoints are equal) and&mdash;potentially&mdash;logic (e.g. check for non-equal constant endpoints at the symbolic level).

PR #89, opened 2021-11-10
Revert "Allow both edges of IntervalTransform to be None"

brandonwillard:
This reverts commit 9298dac59ac45c66912f181d5eb3faf23dc82710.

### DIFF
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -224,8 +224,6 @@ class IntervalTransform(RVTransform):
             return at.log(value - a)
         elif b is not None:
             return at.log(b - value)
-        else:
-            return value
 
     def backward(self, value, *inputs):
         a, b = self.args_fn(*inputs)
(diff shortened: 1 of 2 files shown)


ricardoV94:
I don't see how this makes sense without raising a ValueError. This transform is now inconsistent. Forward and backward will return None whereas log_jac_det will return value when both a and b are None.

That's just wrong

brandonwillard:
> I don't see how this makes sense without raising a ValueError. This transform is now inconsistent. Forward and backward will return None whereas log_jac_det will return value when both a and b are None.
> 
> That's just wrong

What you have just stated is that an inconsistently defined transform is inconsistent.  Yes, that's true, but saying "This transform is now inconsistent." is patently false.

A reversion does not introduce new code&mdash;by definition.  The reversion was made so that we could cut a release with a bug fix and new features without also cutting a release with an addition that **introduces a clear inconsistency** (e.g. the&mdash;at best&mdash;no-op graph clutter caused by allowing two `None` intervals).

We do **not** want a release that quite literally invites people to specify a no-op that only complicates our situation.

Regardless, a `ValueError` can be added at any time.

PR #95, opened 2021-11-19
Use specialized CheckAndRaise Op in logprob graphs

### LINKED ISSUE #84: AssertionError in aeppl.logp breaks sampling in pymc

https://github.com/aesara-devs/aeppl/blob/18275e2bba83f21da5cde256bd67d86fa5b32451/aeppl/logprob.py#L92

**Please provide the full traceback of any errors.**
```python
AssertionError: sigma > 0
Apply node that caused the error: Assert{msg='sigma > 0'}(Elemwise{Composite{((i0 + (i1 * sqr(i2))) - log(i3))}}[(0, 2)].0, All.0)
```

**Please provide any additional information below.**
Sampling with PyMC (master) fails with assertion errors and does not treat AssertionErorr as divergent sample

**Expected Behaviour**
```python
return -inf
```
**Possible Solution**
* provide a graph rewrite for fixing assertions
* replace Assert with switch

This snippet solved my issue
```python
aesara.assert_op.Assert = lambda name: (lambda res, *cond: aesara.tensor.switch(
    aesara.tensor.all(aesara.tensor.stack([c.all() for c in cond])), 
    res, 
    -np.inf
))
```
## Versions and main components

* Aesara version:
* Aesara config (`python -c "import aesara; print(aesara.config)"`)
* Python version:
* Operating system:
* How did you install Aesara: (conda/pip)

ricardoV94:
~The new Op is a bit of a mouthful, so suggestions for a more concise name are welcome.~

~To change the type of Error raised by the Assert Op we would need to override the `perform` and `c_code` methods. This is not difficult, but should we perhaps just make the base Assert Op more flexible on the Aesara side?~

Closes #84

### DIFF
--- a/aeppl/logprob.py
+++ b/aeppl/logprob.py
@@ -4,12 +4,30 @@ from typing import Tuple
 import aesara.tensor as at
 import aesara.tensor.random.basic as arb
 import numpy as np
-from aesara.assert_op import Assert
 from aesara.graph.op import Op
+from aesara.raise_op import CheckAndRaise
 from aesara.tensor.slinalg import Cholesky, solve_lower_triangular
 from aesara.tensor.var import TensorVariable
 
 # from aesara.tensor.xlogx import xlogy0
+
+
+class ParameterValueError(ValueError):
+    """Exception for invalid parameters values in logprob graphs"""
+
+
+class CheckParameterValue(CheckAndRaise):
+    """Implements a parameter value check in a logprob graph.
+
+    Raises `ParameterValueError` if the check is not True.
+    """
+
+    def __init__(self, msg=""):
+        super().__init__(ParameterValueError, msg)
+
+    def __str__(self):
+        return f"Check{{{self.msg}}}"
+
 
 cholesky = Cholesky(lower=True, on_error="nan")
 
@@ -108,7 +126,7 @@ def uniform_logcdf(op, value, *inputs, **kwargs):
         ),
     )
 
-    res = Assert("lower <= upper")(res, at.all(at.le(lower, upper)))
+    res = CheckParameterValue("lower <= upper")(res, at.all(at.le(lower, upper)))
     return res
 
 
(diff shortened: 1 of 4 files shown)


brandonwillard (commented):
I believe we need to raise a custom exception, which means we need to create an entirely new `Op` that is nearly the same as `aesara.assert_op.Assert`, except that its `Op.c_code` implementation raises a custom exception type.  Here's a PR for such an `Op`: https://github.com/aesara-devs/aesara/pull/672.

ricardoV94:
Will need a Aesara release to be able to use the new `CheckAndRais` `Op`

tests/test_printing.py
```diff
@@ -41,18 +41,19 @@ def test_notex_print():
     )
     assert pprint(normalrv_noname_expr) == expected.strip()
 
+    # TODO: This is no longer working with Aesara==2.3.0
```
ricardoV94, line 44: The behavior changed because of https://github.com/aesara-devs/aesara/commit/3d4ef668cafce97acfab095e0c1a44752abefb2c

@brandonwillard Is there an easy fix?

tests/test_utils.py
```diff
@@ -34,7 +34,8 @@ def test_change_rv_size():
     new_size = np.array([4, 3], dtype="int32")
     rv_newer = change_rv_size(rv, new_size=new_size, expand=False)
     assert rv_newer.ndim == 2
-    assert isinstance(rv_newer.owner.inputs[1], Constant)
+    # TODO: This is no longer working with Aesara==2.3.0
```
ricardoV94, line 37: Same commit involved here

tests/test_utils.py
```diff
@@ -34,7 +34,7 @@ def test_change_rv_size():
     new_size = np.array([4, 3], dtype="int32")
     rv_newer = change_rv_size(rv, new_size=new_size, expand=False)
     assert rv_newer.ndim == 2
-    assert isinstance(rv_newer.owner.inputs[1], Constant)
+    assert isinstance(rv_newer.owner.inputs[1].owner.inputs[1], Constant)
```
ricardoV94, line 37: Related, the input for non-scalar sizes is now a SpecifyShape

tests/test_printing.py
brandonwillard: https://github.com/aesara-devs/aesara/pull/680 should fix this specific issue.

tests/test_utils.py
brandonwillard: https://github.com/aesara-devs/aesara/pull/680 should fix this as well.

brandonwillard (approved):

PR #101, opened 2021-12-23
Allow advanced/fancy indexing and arbitrary `Join` axes in mixtures

brandonwillard:
This PR enables support for arbitrary NumPy indexing.  It also allows mixtures that are `Join`ed (e.g. `at.stack`ed) on any axis.  The `Join` axis must be have an appropriate `MeasurableVariable` index, though, and it must be the only index that has one (for now).

Here's an example:

``` python
import numpy as np
import aesara.tensor as at


srng = at.random.RandomStream(29833)

# A "time" dimension
T_at = at.lscalar("T")

X_rv = srng.normal(-100, 1, size=T_at, name="X")
Y_rv = srng.gamma(0.5, 0.5, size=T_at, name="Y")
Z_rv = srng.normal(100, 1, size=T_at, name="Z")

I_rv = srng.categorical(np.r_[0.1, 0.2, 0.7], size=T_at, name="I")

i_vv = I_rv.clone()
i_vv.name = "i"

# Select a random mixture component across the "time" dimension (i.e. the
# first dimension)
M_rv = at.stack([X_rv, Y_rv, Z_rv], axis=1)[at.arange(T_at), I_rv]
M_rv.name = "M"

m_vv = M_rv.clone()
m_vv.name = "m"

logp_parts = joint_logprob({M_rv: m_vv, I_rv: i_vv}, sum=False)

# Evaluate the log-probability at a test point
res = logp_parts.eval(
    {T_at: 5, m_vv: np.r_[-100, -100, 1, 100, -100], i_vv: np.r_[0, 0, 1, 2, 0]}
)
res
# array([-3.22152363, -3.22152363, -3.02837643, -1.27561348, -3.22152363])

exp_res = np.array(
    [scipy.stats.norm(-100, 1).logpdf(-100) + np.log(0.1)] * 2
    + [scipy.stats.gamma(0.5, scale=2.0).logpdf(1) + np.log(0.2)]
    + [scipy.stats.norm(100, 1).logpdf(100) + np.log(0.7)]
    + [scipy.stats.norm(-100, 1).logpdf(-100) + np.log(0.1)]
)
assert np.allclose(exp_res, res)
```


kc611 (approved):

kc611:
We could also calculate the indices by splitting the advanced indices into basic ones and directly stacking them on top of one another (in shape of the subspace generated) in case of moved subspaces. But in the end the result would be the same as this logic.

ricardoV94 (commented):
LGTM. This is a really cool addition!

Otherwise, some branches of the `expand_indices` seem to not be covered by the tests, as well as the tag.test_value logic in `mixture_replace`.

brandonwillard:
> We could also calculate the indices by splitting the advanced indices into basic ones and directly stacking them on top of one another (in shape of the subspace generated) in case of moved subspaces. But in the end the result would be the same as this logic.

I'm interested in this approach as well.  Can you mock up an example of this conversion in plain NumPy?  Would be good to consider it in case it has advantages over this approach.

brandonwillard:
> LGTM. This is a really cool addition!
> 
> Otherwise, some branches of the `expand_indices` seem to not be covered by the tests, as well as the tag.test_value logic in `mixture_replace`.

Yes, I have a bunch of tests to add before we can merge this.  I might also try to add `np.newaxis` support.

kc611:
> Can you mock up an example of this conversion in plain NumPy? Would be good to consider it in case it has advantages over this approach.

Sure, I built these implementations: https://gist.github.com/kc611/8609c7d0e20558a36a7d927b1c5cc065

Seems like the logic gets more complex when we try to expand slices. But it is kind of simple if we treat slices as one basic index. (Made two implementations to reflect that difference)

brandonwillard:
> This implementation (without stacking) can be also be useful in implementation of advanced indexing in Numba since indexing there is currently being done axis-wise not element-wise like we are doing right now. (Currently in Numba for one dimensional advanced indexing, the loops traverse it axis by axis hence they require the array of the indices of the current axis instead of the current element.)

Yeah, it might work well for a reimplementation in Numba when&mdash;say&mdash;the advanced indexing sub-space can be iterated efficiently.  

For the log-probability computations performed here, we need to iterate on the axis that joins the `MeasurableVariable`s, which also requires that we be able to split and filter the indices on each dimension accordingly, and I'm not sure how to do that in the basic index case&mdash;at least not without eventually turning each index into an `ndarray` that can be `np.nonzero`ed (i.e. into an advanced index).

Regardless, this approach is a very good insight.

PR #110, opened 2022-01-12
Add Dirac-delta support

brandonwillard:
This PR adds Dirac-delta support via an ephemeral `DiracDelta` `Op`.  The relevant lifting and removal rewrites have been added as well.

### DIFF
--- a/aeppl/logprob.py
+++ b/aeppl/logprob.py
@@ -9,7 +9,7 @@ from aesara.raise_op import CheckAndRaise
 from aesara.tensor.slinalg import Cholesky, solve_lower_triangular
 from aesara.tensor.var import TensorVariable
 
-# from aesara.tensor.xlogx import xlogy0
+from aeppl.dists import DiracDelta
 
 
 class ParameterValueError(ValueError):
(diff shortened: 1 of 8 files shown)


aeppl/logprob.py
```diff
@@ -596,3 +596,10 @@ def multinomial_logprob(op, values, *inputs, **kwargs):
         at.all(at.ge(n, 0)),
     )
     return res
+
+
+@_logprob.register(DiracDelta)
+def diracdelta_logprob(op, values, *inputs, **kwargs):
+    (values,) = values
+    (const_value,) = inputs
+    return at.switch(at.eq(values, const_value), 0.0, -np.inf)
```
ricardoV94, line 605: Os strict equality or `isclose` best here?

ricardoV94 (commented):
Why not make it a RandomVariable? thay way the default transforms would work out of the box, and we can still have the remove useless rewrite.

brandonwillard:
> Why not make it a RandomVariable? thay way the default transforms would work out of the box, and we can still have the remove useless rewrite.

`RandomVariable` interface simply doesn't fit this type of "distribution" well enough, so there would be a bit more boilerplate to that.  Plus, it would only afford a minor convenience as a result.  Also, this type of `Op` has more simplifying properties that are easy to express directly, whereas `RandomVariable` doesn't, so this `Op` would be unnecessarily restricted by what was available to `RandomVariable`s&mdash;unless we implemented its special properties (e.g. nearly every `Op` can be lifted through it), but then we're already doing more work than the current approach.

aeppl/logprob.py
brandonwillard: Yeah, we could use that.

ricardoV94:
Yeah nevermind. I was thinking only about the Dimshuffle and broadcast lifts, but here you are doing it for every unary operator as well.

ricardoV94 (dismissed):

PR #116, opened 2022-01-31
Do not replace valued RVs in `naive_bcast_rv_lift`

### LINKED ISSUE #115: Broadcasted variable not replaced in downstream terms

```python
import aesara.tensor as at
import aeppl

cov = at.eye(2)
mu_rv = at.random.normal(size=2, name='mu')
mu_broadcasted = at.broadcast_arrays(mu_rv, cov[..., -1])[0]
# mu_broadcasted = mu_rv  # No error if this line is removed
x_rv = at.random.multivariate_normal(mu_broadcasted, cov, name='x')

mu_vv = mu_rv.type()
x_vv = x_rv.type()
logp = aeppl.factorized_joint_logprob({mu_rv: mu_vv, x_rv: x_vv})
```

```python
aeppl/joint_logprob.py:161: UserWarning: Found a random variable that was neither among the observations nor the conditioned variables: normal_rv{0, (0, 0), floatX, False}(RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F51B7F3EC80>), TensorConstant{[]}, TensorConstant{11}, TensorConstant{(2,) of 0.0}, TensorConstant{(2,) of 1.0})
  warnings.warn(
```

```python
logp[x_vv].eval({mu_vv: [0, 0], x_vv: [0, 0]})
```

```python
UnusedInputError                          Traceback (most recent call last)
/tmp/ipykernel_55923/2490478877.py in <module>
----> 1 logp[x_vv].eval({mu_vv: [0, 0], x_vv: [0, 0]})

~/miniconda3/envs/aeppl/lib/python3.9/site-packages/aesara/graph/basic.py in eval(self, inputs_to_values)
    548         inputs = tuple(sorted(inputs_to_values.keys(), key=id))
    549         if inputs not in self._fn_cache:
--> 550             self._fn_cache[inputs] = function(inputs, self)
    551         args = [inputs_to_values[param] for param in inputs]
    552 

~/miniconda3/envs/aeppl/lib/python3.9/site-packages/aesara/compile/function/__init__.py in function(inputs, outputs, mode, updates, givens, no_default_updates, accept_inplace, name, rebuild_strict, allow_input_downcast, profile, on_unused_input)
    335         # note: pfunc will also call orig_function -- orig_function is
    336         #      a choke point that all compilation must pass through
--> 337         fn = pfunc(
    338             params=inputs,
    339             outputs=outputs,

~/miniconda3/envs/aeppl/lib/python3.9/site-packages/aesara/compile/function/pfunc.py in pfunc(params, outputs, mode, updates, givens, no_default_updates, accept_inplace, name, rebuild_strict, allow_input_downcast, profile, on_unused_input, output_keys)
    522         inputs.append(si)
    523 
--> 524     return orig_function(
    525         inputs,
    526         cloned_outputs,

~/miniconda3/envs/aeppl/lib/python3.9/site-packages/aesara/compile/function/types.py in orig_function(inputs, outputs, mode, accept_inplace, name, profile, on_unused_input, output_keys)
   1971     try:
   1972         Maker = getattr(mode, "function_maker", FunctionMaker)
-> 1973         m = Maker(
   1974             inputs,
   1975             outputs,

~/miniconda3/envs/aeppl/lib/python3.9/site-packages/aesara/compile/function/types.py in __init__(self, inputs, outputs, mode, accept_inplace, function_builder, profile, on_unused_input, fgraph, output_keys, name)
   1574 
   1575         # Check if some input variables are unused
-> 1576         self._check_unused_inputs(inputs, outputs, on_unused_input)
   1577 
   1578         # Make a list of (SymbolicInput|SymblicInputKits, indices,

~/miniconda3/envs/aeppl/lib/python3.9/site-packages/aesara/compile/function/types.py in _check_unused_inputs(self, inputs, outputs, on_unused_input)
   1746                     )
   1747                 elif on_unused_input == "raise":
-> 1748                     raise UnusedInputError(msg % (inputs.index(i), i.variable, err_msg))
   1749                 else:
   1750                     raise ValueError(

UnusedInputError: aesara.function was asked to create a function computing outputs given certain inputs, but the provided input variable at index 0 is not part of the computational graph needed to compute the outputs: <TensorType(float64, vector)>.
To make this error into a warning, you can pass the parameter on_unused_input='warn' to aesara.function. To disable it completely, use on_unused_input='ignore'.

### PR DESCRIPTION
**Pull request by ricardoV94**
Closes #115

### DIFF
--- a/aeppl/opt.py
+++ b/aeppl/opt.py
@@ -219,6 +219,13 @@ def naive_bcast_rv_lift(fgraph, node):
     if hasattr(fgraph, "dont_touch_vars") and rv_var in fgraph.dont_touch_vars:
         return None  # pragma: no cover
 
+    # Do not replace RV if it is associated with a value variable
+    rv_map_feature: Optional[PreserveRVMappings] = getattr(
+        fgraph, "preserve_rv_mappings", None
+    )
+    if rv_map_feature is not None and rv_var in rv_map_feature.rv_values:
+        return None
+
     if not bcast_shape:
         # The `BroadcastTo` is broadcasting a scalar to a scalar (i.e. doing nothing)
         assert rv_var.ndim == 0
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,10 +1,5 @@
-import sys
 import os
-import pathlib
 
-
-# import local version of library instead of installed one
-sys.path.insert(0, str(pathlib.Path(__file__).parent.resolve().parent.parent / "src"))
 import aeppl
 
 # -- Project information
--- a/tests/test_opt.py
+++ b/tests/test_opt.py
@@ -1,17 +1,20 @@
 import aesara
 import aesara.tensor as at
+import numpy as np
+import scipy.stats as st
 from aesara.graph.opt import in2out
 from aesara.graph.opt_utils import optimize_graph
 from aesara.tensor.elemwise import DimShuffle, Elemwise
 from aesara.tensor.extra_ops import BroadcastTo
 from aesara.tensor.subtensor import Subtensor
 
+from aeppl import factorized_joint_logprob
 from aeppl.dists import DiracDelta, dirac_delta
 from aeppl.opt import local_lift_DiracDelta, naive_bcast_rv_lift
 
 
 def test_naive_bcast_rv_lift():
-    r"""Make sure `test_naive_bcast_rv_lift` can handle useless scalar `BroadcastTo`\s."""
+    r"""Make sure `naive_bcast_rv_lift` can handle useless scalar `BroadcastTo`\s."""
     X_rv = at.random.normal()
     Z_at = BroadcastTo()(X_rv, ())
 
@@ -20,6 +23,26 @@ def test_naive_bcast_rv_lift():
 
     res = optimize_graph(Z_at, custom_opt=in2out(naive_bcast_rv_lift), clone=False)
     assert res is X_rv
+
+
+def test_naive_bcast_rv_lift_valued_var():
+    r"""Check that `naive_bcast_rv_lift` won't touch valued variables"""
+
+    x_rv = at.random.normal(name="x")
+    broadcasted_x_rv = at.broadcast_to(x_rv, (2,))
+
+    y_rv = at.random.normal(broadcasted_x_rv, name="y")
+
+    x_vv = x_rv.clone()
+    y_vv = y_rv.clone()
+    logp_map = factorized_joint_logprob({x_rv: x_vv, y_rv: y_vv})
+    assert x_vv in logp_map
+    assert y_vv in logp_map
+    assert len(logp_map) == 2
+    assert np.allclose(logp_map[x_vv].eval({x_vv: 0}), st.norm(0).logpdf(0))
+    assert np.allclose(
+        logp_map[y_vv].eval({x_vv: 0, y_vv: [0, 0]}), st.norm(0).logpdf([0, 0])
+    )
 
 
 def test_local_lift_DiracDelta():

### CODE REVIEW

**Review by brandonwillard — CHANGES_REQUESTED** [review]
It seems like this is trying to prevent rewrites on valued/observed random variables, but we definitely don't want to do that.  

If relationships between rewritten terms aren't being preserved properly, then it sounds like there's an issue with the rewriting process.

**Comment by ricardoV94** [comment]
The more I think about this rewrite, the less convinced I am it ever makes sense. If you have `x_rv` and `at_broadcast(x_rv, shape)` as the outputs of a function, the actual values may become almost independent (subject to intricacies of the rng) once this rewrite is applied to the second node (even if there is no resizing). 

For me, this only really makes sense to facilitate uncovering rv-valued variables.

**Comment by brandonwillard** [comment]
> The more I think about this rewrite, the less convinced I am it ever makes sense. If you have `x_rv` and `at_broadcast(x_rv, shape)` as the outputs of a function, the actual values may become almost independent (subject to intricacies of the rng) once this rewrite is applied to the second variable.
> 
> For me, this only really makes sense to facilitate uncovering rv-valued variables.

It sounds like you're referring to the representation issue addressed in https://github.com/aesara-devs/aeppl/pull/78.

The `RandomVariable`s produced after lifting the broadcast operations aren't "proper" random variables.  The entire purpose of this rewriting is to produce a log-probability graph, and those rewritten `RandomVariable`s only serve as type-based measure information that we use in order to determine/derive the log-probabilities.  

Simply put, when a `BroadcastTo` is lifted through a `RandomVariable`, we're not trying to produce the equivalent `RandomVariable`; we're constructing a `RandomVariable` of the same type, and with properly broadcastated parameters, that **when evaluated at a given point will produce the same log-probability as the un-lifted `RandomVariable` would.**

The graphs that make up our IR are never meant to be sampled, so RNG-based (in)dependence isn't relevant.  For that matter, RNGs themselves aren't relevant to these graphs.  If you're thinking of a rewrite that's valid in sample-space graphs, then you're not considering the space we're actually working within.

In other words, while two random variables may be distinct, their measures can still be equivalent, and that's what we're dealing with here.  The confusion mostly arises from us not making/having a very formal distinction between our graphical representations of the two scenarios.


https://github.com/aesara-devs/aeppl/pull/78 addresses some of this by making such rewrites applicable only to distinct valued `RandomVariable` objects.  We could also create replacement `Op`s for `RandomVariable`s that&mdash&say&mdash;remove the RNG components and have a more revealing name (e.g. something having to do with them providing measure information), but I don't know how useful that would really be.

**Comment by brandonwillard** [comment]
> This feels very specific to RandomVariables which are sort of an "input" variable and I don't think would be a problem with the other type of rewrites we are doing (although we have been limiting it to nodes that have corresponding value variables, so maybe it is a more general issue)

This sounds like a line of investigation that will lead to the real problem.

**Comment by brandonwillard** [comment]
If you rebase this onto `main`, the tests should run.

**Comment by ricardoV94** [comment]
> This sounds like a line of investigation that will lead to the real problem.

I understand the real problem. I am not sure what type of solution you have in mind.

This change is not affecting our ability to parse logprob terms because you couldn't ever measure both the broadcasted node and the base rv at the same time anyway. 

If you are trying to measure the broadcasted node it will still work, and if you are trying to measure the base RV node and only then broadcast the value this won't mess up the graph dependencies. In both cases the behavior is what we want for the IR.

Third case is when you have several RandomVariables that you want to keep in the logprob graph and you have clients for both the base and broadcasted nodes. That would be the "random graph" issue I mentioned above. The solution is to limit this rewrite even more than what I did here and only allow it for valued broadcasted nodes.

**Review by brandonwillard — COMMENTED** [review]
What if you added something like the following to `naive_bcast_rv_lift`:
```python
    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)

    if rv_map_feature is not None and rv_var in rv_map_feature.rv_values:
        val_var = rv_map_feature.rv_values[rv_var]
        val_var = at.broadcast_to(val_var, tuple(bcast_shape))
        rv_map_feature.update_rv_maps(rv_var, val_var, bcasted_node.outputs[1])
```

The main problem I see here is that `naive_bcast_rv_lift` simply isn't value variable conscientious like the other rewrites are.

PR #121, opened 2022-02-08
Track value variables when lifting `BroadcastTo` `Op`s

### LINKED ISSUE #115: Broadcasted variable not replaced in downstream terms

brandonwillard:
This PR is an alternative to #116 that enables value-variable tracking for `BroadcastTo` lifting rewrites.  This approach allows one to specify values for broadcasted random variables that are also valued, whereas globally disabling the rewrite for valued terms would not.

The latter is most easily demonstrated by the following example:
``` python
import aesara.tensor as at

X_rv = at.random.normal(name="X")
Z_rv = at.broadcast_to(X_rv, (2, 2))

x_vv = X_rv.clone()
z_vv = Z_rv.clone()

logp_map = factorized_joint_logprob({X_rv: x_vv, Z_rv: z_vv})
```

Closes #115

Also, this PR further demonstrates the relevance of #78, because it explicitly makes use of the fact that the lifting operation doesn't only apply to the valued variable, but also the variable's value.

### DIFF
--- a/aeppl/opt.py
+++ b/aeppl/opt.py
@@ -244,10 +244,20 @@ def naive_bcast_rv_lift(fgraph, node):
     ]
     bcasted_node = lifted_node.op.make_node(rng, size, dtype, *new_dist_params)
 
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    new_bcast_out = bcasted_node.outputs[1]
+
+    if rv_map_feature is not None and rv_var in rv_map_feature.rv_values:
+        val_var = rv_map_feature.rv_values[rv_var]
+        new_val_var = at.broadcast_to(val_var, tuple(bcast_shape))
+        rv_map_feature.rv_values[new_bcast_out] = new_val_var
+        rv_map_feature.original_values[new_val_var] = new_val_var
+
     if aesara.config.compute_test_value != "off":
         compute_test_value(bcasted_node)
 
-    return [bcasted_node.outputs[1]]
+    return [new_bcast_out]
 
 
 logprob_rewrites_db = SequenceDB()
--- a/tests/test_opt.py
+++ b/tests/test_opt.py
@@ -1,17 +1,20 @@
 import aesara
 import aesara.tensor as at
+import numpy as np
+import scipy.stats as st
 from aesara.graph.opt import in2out
 from aesara.graph.opt_utils import optimize_graph
 from aesara.tensor.elemwise import DimShuffle, Elemwise
 from aesara.tensor.extra_ops import BroadcastTo
 from aesara.tensor.subtensor import Subtensor
 
+from aeppl import factorized_joint_logprob
 from aeppl.dists import DiracDelta, dirac_delta
 from aeppl.opt import local_lift_DiracDelta, naive_bcast_rv_lift
 
 
 def test_naive_bcast_rv_lift():
-    r"""Make sure `test_naive_bcast_rv_lift` can handle useless scalar `BroadcastTo`\s."""
+    r"""Make sure `naive_bcast_rv_lift` can handle useless scalar `BroadcastTo`\s."""
     X_rv = at.random.normal()
     Z_at = BroadcastTo()(X_rv, ())
 
@@ -20,6 +23,39 @@ def test_naive_bcast_rv_lift():
 
     res = optimize_graph(Z_at, custom_opt=in2out(naive_bcast_rv_lift), clone=False)
     assert res is X_rv
+
+
+def test_naive_bcast_rv_lift_valued_var():
+    r"""Check that `naive_bcast_rv_lift` handles valued variables correctly."""
+
+    x_rv = at.random.normal(name="x")
+
+    mu = at.broadcast_to(x_rv, (2,))
+    y_rv = at.random.normal(mu, name="y")
+
+    x_vv = x_rv.clone()
+    y_vv = y_rv.clone()
+    logp_map = factorized_joint_logprob({x_rv: x_vv, y_rv: y_vv})
+
+    assert x_vv in logp_map
+    assert y_vv in logp_map
+    y_val = np.array([0, 0])
+    assert np.allclose(
+        logp_map[y_vv].eval({x_vv: 0, y_vv: y_val}), st.norm(0).logpdf(y_val)
+    )
+
+    # Lifting should also work when `BroadcastTo`s are directly assigned value
+    # variables
+    z_rv = at.broadcast_to(x_rv, (2, 2))
+    z_rv.name = "Z"
+    z_vv = z_rv.clone()
+    z_vv.name = "z"
+
+    logp_map = factorized_joint_logprob({x_rv: x_vv, z_rv: z_vv})
+    assert x_vv in logp_map
+    assert z_vv in logp_map
+    z_val = np.array([[0, 0], [0, 0]])
+    assert np.allclose(logp_map[z_vv].eval({z_vv: z_val}), st.norm(0).logpdf(z_val))
 
 
 def test_local_lift_DiracDelta():
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,10 +1,5 @@
-import sys
 import os
-import pathlib
 
-
-# import local version of library instead of installed one
-sys.path.insert(0, str(pathlib.Path(__file__).parent.resolve().parent.parent / "src"))
 import aeppl
 
 # -- Project information


tests/test_opt.py
```diff
@@ -22,6 +25,39 @@ def test_naive_bcast_rv_lift():
...
+    logp_map = factorized_joint_logprob({x_rv: x_vv, y_rv: y_vv})
+
+    assert x_vv in logp_map
+    assert y_vv in logp_map
+    y_val = np.array([0, 0])
+    assert np.allclose(
+        logp_map[y_vv].eval({x_vv: 0, y_vv: y_val}), st.norm(0).logpdf(y_val)
+    )
+
+    # Lifting should also work when `BroadcastTo`s are directly assigned value
```
ricardoV94, line 47: I don't think this example should be valid. We are measuring twice the same thing (pre-broadcast and post broadcasted value variable)

Besides, this would be fine with saying `{x_vv: 0, z_vv: [1, 1]}` which does not match the original graph

ricardoV94: If either only one of the pre- and post- broadcasted variable is valued then #116 already worked.

... although the post-broadcast case is still a bit iffy and depends on what we mean by joint-logprob. The original generative graph of `z_rv` would correspond to a logp of something like:

```
def broadcastedTo_logp(value, original_shape, dist_op, dist_params):

  # e.g., undo_broadcast((3, 3, 3), (1,)) -> (3,), True
  pre_broadcast_value, possible = undo_broadcast(value, original_shape)
  
  return switch(
      possible,
      logprob(dist_op, pre_broadcast_value, *dist_params),
      -inf,
      )
  )
```

aeppl/opt.py
```diff
@@ -244,10 +244,20 @@ def naive_bcast_rv_lift(fgraph, node):
...
     bcasted_node = lifted_node.op.make_node(rng, size, dtype, *new_dist_params)
 
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    new_bcast_out = bcasted_node.outputs[1]
+
+    if rv_map_feature is not None and rv_var in rv_map_feature.rv_values:
+        val_var = rv_map_feature.rv_values[rv_var]
+        new_val_var = at.broadcast_to(val_var, tuple(bcast_shape))
+        rv_map_feature.rv_values[new_bcast_out] = new_val_var
```
ricardoV94, line 254: Won't this add an extra logp term? If I am parsing this correctly, in the `joint_logp` of your first new test case there would now be 3 terms corresponding to `(x_vv, new_value_var, y_vv)`, no?

tests/test_opt.py
brandonwillard: > If either only one of the pre- and post- broadcasted variable is valued then #116 already worked.

Only if it broadcasted something that wasn't valued, but that's the limitation I've demonstrated. 

> ... although the post-broadcast case is still a bit iffy and depends on what we mean by joint-logprob. Strictly speaking, the original generative graph of `z_rv` would correspond to a logp of something like:
> 
> ```python
> def broadcastTo_logp(value, original_shape, dist_op, dist_params):
> 
>   # e.g., undo_broadcast((3, 3, 3), (1,)) -> (3,)
>   pre_broadcast_value = undo_broadcast(value, original_shape)
>   
>   return switch(
>     at.eq(value, at.broadcast_to(pre_broadcast_value, value.shape)),
>     logprob(dist_op, pre_broadcast_value, *dist_params),
>     -inf,
>   )  
> ```
> 
> But that's a separate question from the comment above

You're goint to need to clarify this, because, right now, I don't see the relevance of this function.  Can you demonstrate the underlying issue by deriving an incorrect log-probability calculation under these changes?

aeppl/opt.py
brandonwillard: > Won't this add an extra logp term? If I am parsing this correctly, calling `joint_logp` on your first new test case would now sum 3 terms corresponding to `(x_vv, new_value_var, y_vv)`, no?

Have you tried it?  We can always check the sum in the tests.

brandonwillard: The first test case might actually add an extra term; I'll try it out.

brandonwillard:
We would need to do something about the way value variables are used in the `factorized_joint_logprob` logic in order to remove the extra term @ricardoV94 mentioned, but this is essentially what #78 is trying to do.

Aside from that, we can have the restriction of not computing log-probability graphs for transformed valued variables that already appear in the same call to `*joint_logprob`, but we should do something about what it currently handles that (i.e. simply ignoring the transformed graph).

ricardoV94:
That seems somewhat related to https://github.com/aesara-devs/aeppl/discussions/119 as well btw

PR #128, opened 2022-03-21
Rename `Simplex` to `SimplexTransform` for consistency with other transforms

ricardoV94:


twiecki:
Is there no test for this transform?

ricardoV94:
> Is there no test for this transform?

Should be tested here: https://github.com/aesara-devs/aeppl/blob/751979802f1aef5478fdbf7cc1839df07df60825/tests/test_transforms.py#L145-L160

twiecki:
I see, I was just wondering why there was no change to the test, but it makes sense now.

twiecki (approved):

Issue #85, opened 2021-11-09
API of aeppl.factorized_joint_logprob


aseyboldt:
During discussion of https://github.com/pymc-devs/pymc/issues/5155 we (@ricardoV94 and I) also talked a bit about the API for `aeppl.factorized_joint_logprob` and how it deals with random variables that are not specified at all.

I think it might be better to change the api a bit so that all random variables in the graph have to be specified somehow. That would make it more future proof (eg implement marginalization using closed form solution or some kind of numerical integration) and also make it closer to the math notation and harder for users to shoot themselves into the foot.

Let's say we have a graph like this:
```python
import aesara
import aesara.tensor as at
import aeppl

a = at.random.normal()
b = at.random.normal(loc=a)

a_val = at.dscalar()
b_val = at.dscalar()
```

We could now be interested in those values (math notation):

```
# joint logp
P(a = a_val, b = b_val)

# some conditional logps
P(a = a_val | b = b_val) = P(a = a_val)
P(b = b_val | a = a_val)

# marginal logps
P(a = a_val) = \int_b_val P(a = a_val | b = b_val) P(b = b_val)  # integral doesn't do anything...
P(b = b_val) = \int_a_val P(b = b_val | a = a_val) P(b = b_val)  # this one does

# Random variables whose expectation is the marginal logp (but that might converge *really* slowly or even not at all in many cases)
P_rv(a = a_val | b = B), where B ~ b  # not sure actually how to write this down properly
P_rv(b = b_val | a = A), where A ~ a
```

If we call `aeppl.factorized_joint_logprob(all_vars)`, we get all the conditional logps: `{key_var: P(key_var = key_val | all_remaining) for (key_var, key_val) in all_vars}`, which seems fine to me (although I might only be interested in some of those in the first place, but I think this is fine anyway).

If we call `aeppl.factorized_joint_logprob({b: b_val})` however, we actually get the random variable instead of the marginal logp, even though the call looks almost like the math notation of the marginal logp.

We could make this choice explicit by switching this use case to an api like this:
```python
aeppl.factorized_joint_logprob({b: b_val}, sample=[a])
```

Which we could later extend to marginalization using something those:
```python
aeppl.factorized_joint_logprob({b: b_val}, marginalize=[a])  # closed form solution only
aeppl.factorized_joint_logprob({b: b_val}, marginalize=[a], integration_options={a: "gauss_hermite(21)"}) # some numerical integration
```

brandonwillard:
[...]

Issue #90, opened 2021-11-10
Raise NotImplementedError when factorized_joint_logprob cannot retrieve terms for every value variable


ricardoV94:
This is a follow up on a discussion that started in https://github.com/aesara-devs/aeppl/pull/22#discussion_r736279307

I believe that whenever aeppl cannot provide a rewrite for a given graph the safest way to proceed is to raise an exception. In the following snippet the returned output is not a "joint_lopgrob" graph at all. The slightly obscure warning that is raised right now sounds insufficient. Users should not be expected to know what kind of graphs aeppl will be able to parse or not, and have to investigate the graph to see if some terms happen to be missing.

```python
import aesara.tensor as at
import aeppl

x_rv = at.random.normal(name="x")
y_rv = at.cos(x_rv)

z_rv = at.random.normal(name='z')  # Just so that the returned output is not `None`

y_vv = y_rv.clone()
z_vv = z_rv.clone()

aeppl.joint_logprob({y_rv: y_vv, z_rv: z_vv})
```
```
UserWarning: Found a random variable that was neither among the observations nor the conditioned variables: normal_rv{0, (0, 0), floatX, False}(RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FDFC27D6D60>), TensorConstant{[]}, TensorConstant{11}, TensorConstant{0.0}, TensorConstant{1.0})
```
I suggested in that PR to add a simple check at the end of `factorized_joint_logprob` of the form of:

```python
missing_value_terms = set(original_values.values()) - set(logprob_vars.keys()) 
if missing_value_terms:
    raise NotImplementedError(
        f"The logprob terms for the following value variables could not be derived: {missing_value_terms}"
    )
```

brandonwillard:
> The slightly obscure warning that is raised right now sounds insufficient.

Let's start with exactly why that warning is insufficient and how you want to address those inefficiencies.

> Users should not be expected to know what kind of graphs aeppl will be able to parse or not, and have to investigate the graph to see if some terms happen to be missing.

They aren't.  The only thing users are expected to know is the set of terms in their model graphs for which they want to construct log-probability graphs.  That's it.

Also, nothing about improved error messages would require them to know less about the graphs with which they're working.  If a user received more/better error messages from `factorized_joint_logprob`, those messages would only help the user better understand their graphs, but it wouldn't prevent them from needing to read or understand the content of the error messages in order to successfully use AePPL.

[...]

PR #129, opened 2022-03-25
Perform log-probability rewrites on un-valued terms

ricardoV94:
Alternative approach to #102, this time relying on permissive intermediate rewrites whenever the relevant inputs are measurable and not themselves valued.

I think this approach is reasonable, except for the case where we have RVs in the graph that are not meant to be measured (see new xfail test):

```python
import aesara.tensor as at

x = at.random.normal()
x.tag.ignore_logprob = True

y = at.clip(x, 0, 1)

z_rv = at.random.normal(y, 1, name="z")
z_vv = z_rv.clone()

logp = joint_logprob({z_rv: z_vv})
```
```
/home/ricardo/Documents/Projects/aeppl/aeppl/joint_logprob.py:161: UserWarning: Found a random variable that was neither among the observations nor the conditioned variables: [Elemwise{clip}.0]
  warnings.warn(
```

That issue is related to #85

### DIFF
--- a/aeppl/mixture.py
+++ b/aeppl/mixture.py
@@ -5,7 +5,11 @@ import aesara.tensor as at
 from aesara.graph.basic import Apply, Variable
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op, compute_test_value
-from aesara.graph.opt import local_optimizer, out2in, pre_greedy_local_optimizer
+from aesara.graph.opt import (
+    EquilibriumOptimizer,
+    local_optimizer,
+    pre_greedy_local_optimizer,
+)
 from aesara.ifelse import ifelse
 from aesara.tensor.basic import Join, MakeVector
 from aesara.tensor.random.opt import local_dimshuffle_rv_lift, local_subtensor_rv_lift
@@ -371,7 +375,9 @@ def logprob_MixtureRV(
 
 logprob_rewrites_db.register(
     "mixture_replace",
-    out2in(mixture_replace, name="mixture_replace", ignore_newtrees=True),
+    EquilibriumOptimizer(
+        [mixture_replace], max_use_ratio=aesara.config.optdb__max_use_ratio
+    ),
     0,
     "basic",
     "mixture",
--- a/aeppl/mixture.py
+++ b/aeppl/mixture.py
@@ -260,12 +260,9 @@ def mixture_replace(fgraph, node):
 
     old_mixture_rv = node.default_output()
 
-    if old_mixture_rv not in rv_map_feature.rv_values:
-        return None  # pragma: no cover
-
     mixture_res, join_axis = get_stack_mixture_vars(node)
 
-    if mixture_res is None:
+    if mixture_res is None or any(rv in rv_map_feature.rv_values for rv in mixture_res):
         return None  # pragma: no cover
 
     mixing_indices = node.inputs[1:]
@@ -274,10 +271,6 @@ def mixture_replace(fgraph, node):
     # that belong to each one (by way of their indices).
     mixture_rvs = []
     for i, component_rv in enumerate(mixture_res):
-        if component_rv in rv_map_feature.rv_values:
-            raise ValueError(
-                f"A value variable was specified for a mixture component: {component_rv}"
-            )
 
         # We create custom types for the mixture components and assign them
         # null `get_measurable_outputs` dispatches so that they aren't
--- a/tests/test_composite_logprob.py
+++ b/tests/test_composite_logprob.py
@@ -1,0 +1,103 @@
+import aesara
+import aesara.tensor as at
+import numpy as np
+import scipy.stats as st
+
+from aeppl import joint_logprob
+from aeppl.opt import construct_ir_fgraph
+from aeppl.truncation import CensoredRV
+
+
+def test_scalar_clipped_mixture():
+    x = at.clip(at.random.normal(loc=1), 0.5, 1.5)
+    x.name = "x"
+    y = at.random.beta(1, 2, name="y")
+
+    comps = at.stack(x, y)
+    comps.name = "comps"
+    idxs = at.random.bernoulli(0.4, name="idxs")
+    mix = comps[idxs]
+    mix.name = "mix"
+
+    mix_vv = mix.clone()
+    mix_vv.name = "mix_val"
+    idxs_vv = idxs.clone()
+    idxs_vv.name = "idxs_val"
+
+    logp = joint_logprob({idxs: idxs_vv, mix: mix_vv})
+
+    logp_fn = aesara.function([idxs_vv, mix_vv], logp)
+    assert logp_fn(0, 0.4) == -np.inf
+    assert np.isclose(logp_fn(0, 0.5), st.norm.logcdf(0.5, 1) + np.log(0.6))
+    assert np.isclose(logp_fn(0, 1.3), st.norm.logpdf(1.3, 1) + np.log(0.6))
+    assert np.isclose(logp_fn(1, 0.4), st.beta.logpdf(0.4, 1, 2) + np.log(0.4))
+
+
+def test_nested_scalar_mixtures():
+    x = at.random.normal(loc=-50, name="x")
+    y = at.random.normal(loc=50, name="y")
+    comps1 = at.stack(x, y)
+    comps1.name = "comps1"
+    idxs1 = at.random.bernoulli(0.5, name="idxs1")
+    mix1 = comps1[idxs1]
+    mix1.name = "mix1"
+
+    w = at.random.normal(loc=-100, name="w")
+    z = at.random.normal(loc=100, name="z")
+    comps2 = at.stack(w, z)
+    comps2.name = "comps2"
+    idxs2 = at.random.bernoulli(0.5, name="idxs2")
+    mix2 = comps2[idxs2]
+    mix2.name = "mix2"
+
+    comps12 = at.stack(mix1, mix2)
+    comps12.name = "comps12"
+    idxs12 = at.random.bernoulli(0.5, name="idxs12")
+    mix12 = comps12[idxs12]
+    mix12.name = "mix12"
+
+    idxs1_vv = idxs1.clone()
+    idxs2_vv = idxs2.clone()
+    idxs12_vv = idxs12.clone()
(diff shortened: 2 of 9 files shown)


aeppl/mixture.py
```diff
@@ -270,10 +268,6 @@ def mixture_replace(fgraph, node):
     # that belong to each one (by way of their indices).
     mixture_rvs = []
     for i, component_rv in enumerate(mixture_res):
-        if component_rv in rv_map_feature.rv_values:
-            raise ValueError(
-                f"A value variable was specified for a mixture component: {component_rv}"
-            )
```
ricardoV94, line 276: It doesn't make sense to raise, because this could just be an non-valued stack of valued RVs. We could check that the mixture_rv is valued here, but that seems pretty arbitrary in the context of composite logprobs. If this mixture is just a measurable input to another valued RV (and not itself valued), this would not raise, and the user would not notice something is amiss.

I am convinced the best solution is to raise explicitly in `factorized_joint_logprob`, as discussed in #90, and not in the local rewrites.

tests/test_composite_logprob.py
```diff
@@ -0,0 +1,97 @@
...
+    x = at.random.normal()
+    x.tag.ignore_logprob = True
+    y = at.clip(x, 0, 1)
+    z_rv = at.random.normal(y, 1, name="z")
+    z_vv = z_rv.clone()
+
+    logp = joint_logprob({z_rv: z_vv})
+
+    assert not any(
+        isinstance(node.owner.op, CensoredRV)
```
ricardoV94, line 94: This is not an issue in this case as the `CensoredRV` `Op` is just a subclass of the original `Clip` `Op`, but this may not always be the case for our IR, and could become unnecessarily costly as we add more support for basic `Op`s to the limit we might be recreating the whole graph of deterministics with "measurable" equivalents.

brandonwillard: This is the issue I was referring to (indirectly) [here](https://matrix.to/#/!RcUetDjlPGYsTqnQDE:gitter.im/$Ct4DgOBWJHqo7kVWpI2X0JCSHO7hiKVv8Ll_FbZ0BRU?via=gitter.im&via=matrix.org).

My first thought was that we could remove/undo unvalued measurable transforms during some "post-processing" pass of the IR rewrites, or during/after construction of the log-probability graphs.  The latter sounds better, because then we would have much simpler "measurable" IR graphs; ones that could be used for other things that require these "non-computable" measurable `Op`s.

Your comment raises another important point: we should try to standardize the measurable `Op`s&mdash;even the ones that are simple views/clones/wrappers of other `Op`s.  One of the constraints is that such measurable `Op`s essentially act like views/clones of the original `Op`s/`Apply` nodes but without introducing unnecessary indirection (e.g. like using the actual `View` `Op` would).  The subclassing approach seems to work well enough, so we can probably stick with that; however, we need a clean, unified interface for it.

At least one part of this standardization could be addressed through extensions of the `aeppl.abstract.assign_custom_measurable_outputs` constructor function approach.  For instance, we could create a `Variable`-based interface for that constructor that handles all [these steps](https://github.com/aesara-devs/aeppl/blob/a97a6bcbd60726bb5142b39c4da52c7990d50311/aeppl/mixture.py#L281-L283) and only use that constructor wherever we need to convert an existing `Op` to its measurable IR.

brandonwillard (commented):
This looks viable; let's make some refinements and push it through!  

Those refinements really only revolve around the issue(s) demonstrated by `test_overeager_rewrites`, so there might not be much more involved.

brandonwillard:
Just to summarize, we're trying to balance a few things with our measurable IR:
- Ease of use with respect to the terms needed to computed a log-probability
  - Since we need to walk some sub-graphs in order to determine whether or not they're measurable, it helps to collect the relevant terms and make them the inputs of a new, measurable `Op` (e.g. `MixtureRV`)
- Make sure `isinstance(op, MeasurableVariable)` true
  - This pretty fairly straightforward, and there's not much more to do than create/use a type that's been registered as a `MeasurableVariable`
  - This requirement is somewhat duplicated/shadowed by the `_get_measurable_outputs` generic function, which is used to filter out some of the un-measurable outputs of `Scan`s in `aeppl.joint_logprob`.  This implies that our notion of "measurability" **actually** works at the variable/output level and not simply the `Op` level.  **We should clean up the `MeasurableVariable` interface so that it more explicitly represents something like a `MeasurableOp` with a mandatory "measurable output" attribute indicating which outputs of the `Op` are measurable.  Since we already dynamically construct measurable types, we could easily add such a class attribute and possibly remove `_get_measurable_outputs` altogether.**
- Maintain graph consistency with respect to the original un-measured/valued variables' `Op`s and their measurable IR
  - The types, shape inference, and `Op.perform` calculations for un-valued variables in measurable form should not differ
  - Most/all of the consistency issues could be avoided by&mdash;say&mdash;reverting back to the original un-measurable `Op`s when the relevant variables are not actually measured/valued.  In other words, if the variables that aren't measured appear as-is/unchanged in the resulting log-probability graphs, then graph consistency isn't an issue.

ricardoV94:
One note. The only cases I can think of eager rewrites are when we have graphs with RandomVariables (the only `Op`s that are measurable by default), that are not valued or meant to be used as the base of a derived value variable.

If we identified or forced users to mark such variables (e.g., with `ignore_rvs: set`, as mentioned in #) we could make their outputs non-measurable at the onset, which would prevent any eager automatic rewrites which always check if a needed input is measurable

tests/test_composite_logprob.py
ricardoV94: Yes I think we can easily automate most of those steps. In many cases the original Op / node have all the info needed for the logprob function. Some cases that need special inputs like Mixtures perhaps (which can have different original forms), can have custom defined classes as IR but I think those will be the exception rather than the norm.

tests/test_composite_logprob.py
```diff
@@ -0,0 +1,97 @@
...
+    assert np.isclose(logp_fn(0, 0, 0, 50), st.norm.logpdf(100) + np.log(0.5) * 3)
+    assert np.isclose(logp_fn(0, 0, 1, 50), st.norm.logpdf(150) + np.log(0.5) * 3)
+
+
+@pytest.mark.xfail(
+    reason="Measurable rewrites are too eager, when we have a RV in the graph that should not be measured"
+)
+def test_overeager_rewrites():
+    x = at.random.normal()
+    x.tag.ignore_logprob = True
```
brandonwillard, line 86: We still need to remove this `ignore_logprob` feature, and, since nothing depends on it, we should be able to do that now.

brandonwillard (commented):
Let's update the title and description of this PR to reflect its goals and changes, because this is a viable approach.

Aside from that, here are two things we need to address before merging.

### Independent Unit Tests
It would be ideal to have tests for this functionality that don't rely on the mixture and/or censoring rewrites, when that's reasonable.  We just want issues with the composite/nested rewriting to be clearly identifiable and independent of potential issues and changes in our other rewrites.

For the censoring rewrite, that might be one of the simplest nesting examples possible, in which case, using it for our composition tests is probably fine; otherwise, if we plan to extend the censoring rewriting, let's copy its current, simple form and use that copy for testing.

Even so, we still want tests for the interaction between composition and censoring/mixtures/etc., we just don't want to solely rely on those for our composition testing&mdash;at least not if we can avoid it.

### Non-valued IR
Aside from that, we still need to address the IR (intermediate representation) `Op`s that remain in the graph after rewriting&mdash;as demonstrated by `test_overeager_rewrites` (which should be renamed, since "over-eagerness" isn't really the/an underlying issue).

As mentioned in 3. of [this comment](https://github.com/aesara-devs/aeppl/pull/129#issuecomment-1079518290), our "measurable" IR **should** be one-to-one (in terms of computation, at the very least) with its original representation.  This means that we can return these graphs to users&mdash;as this PR currently does; we just need to be clear that we're doing this.

One downside to this approach is that a user compiling such graphs might not get the same performance from our IR as the original, because&mdash;say&mdash;the original sub-graphs have performance-related rewrites that do not apply to their new representations.  To avoid this, I still believe that the `OpFromGraph` approach&mdash;and its corresponding in-lining rewrite&mdash;mentioned in [3.](https://github.com/aesara-devs/aeppl/pull/129#issuecomment-1079518290) is the easiest way to avoid the discrepancy.

At the moment, it looks like `MixtureRV` is the only IR `Op` that removes the original `Op`(s) (i.e. `Join` and/or `MakeVector`), so that's the only existing case that would need to be refactored to use `OpFromGraph`.  The other IR `Op`s are copies/subclasses of the original ones, so most/all rewrites should still apply to them.

Unless I've missed another important "lossy" IR case, we could go ahead with this and follow up on `MixtureRV` later.

brandonwillard (commented):
I've added some refactoring that helps isolate and clarify the measurable IR-specific rewrites.

Also, I made it so that `PreserveRVMappings` tracks un-valued, un-measurable variables that are converted to measurable variables.  This allows us to revert the un-valued measurable variables back into their original un-measurable forms.

I don't particularly like this reversion approach, but it could work for now.

brandonwillard (approved):

ricardoV94:
@brandonwillard I was thinking as an alternative to do a first pass on the graph and convert initial RVs / Measurable Variables that we know aren't needed for the rewrite to be "unmeasurable". This would avoid all cases of unnecessary rewrites I could think of. 

I think this could be done with a single graph walk. The only measurable variables that should remain are those that are directly valued or the first measurable ancestors of valued variables that are not themselves measurable.

ricardoV94:
I think I prefer this approach though. Most cases will not have unused RVs

brandonwillard:
> @brandonwillard, as an alternative I was considering to do a first pass on the graph and convert initial RVs / Measurable Variables that we know aren't needed for the rewrite to be "unmeasurable". This would avoid all cases of unnecessary rewrites I could think of.
> 
> I think this could be done with a single graph walk. The only measurable variables that should remain are those that are directly valued or the first measurable ancestors of valued variables that are not themselves measurable.

For the amount of effort that sounds like it would require, one might as well take one of the more direct approaches proposed early on: e.g. create "guided" rewrites that are aware of the measurability requirements of other (potentially) measurable `Op`'s inputs, and walk/rewrite those inputs.

PR #131, opened 2022-04-06
Implement Truncated RVs

ricardoV94:
**Thank you for opening a PR!**

Here are a few important guidelines and requirements to check before your PR can be merged:
+ [ ] There is an informative high-level description of the changes.
+ [ ] The description and/or commit message(s) references the relevant GitHub issue(s).
+ [ ] [`pre-commit`](https://pre-commit.com/#installation) is installed and [set up](https://pre-commit.com/#3-install-the-git-hook-scripts).
+ [ ] The commit messages follow [these guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
+ [ ] The commits correspond to [_relevant logical changes_](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes), and there are **no commits that fix changes introduced by other commits in the same branch/BR**.  If your commit description starts with "Fix...", then you're probably making this mistake.
+ [ ] There are tests covering the changes introduced in the PR.

Don't worry, your PR doesn't need to be in perfect order to submit it.  As development progresses and/or reviewers request changes, you can always [rewrite the history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_rewriting_history) of your feature/PR branches.

If your PR is an ongoing effort and you would like to involve us in the process, simply make it a [draft PR](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests).


aeppl/truncation.py
```diff
@@ -123,3 +131,127 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+    *rv_inputs_, lower_, upper_ = graph_inputs_
+
+    try:
+        # Use specialized Op
+        truncated_rv_ = _truncated(rv.owner.op, lower_, upper_, *rv_inputs_)
+        return TruncatedRV(
+            inputs=graph_inputs_,
+            outputs=truncated_rv_.owner.outputs,
+            inline=True,
+        )(*graph_inputs)
```
ricardoV94, line 190: Probably it doesn't make sense to wrap these in a `TruncatedRV`, they are supposed to be fully fledged `Ops` with their own `logp` implementations. 

Here, I intended to use the `truncexpon` which is the only truncated implementation in `Aesara` but that one only allows for truncation from above. The example I added here for test with the `uniform` is arguably silly. Other cases would be PyMC's `truncated_normal` which wraps around Scipy's implementation.

brandonwillard: We might need some sort of dispatch for `Op` truncation cases with existing/efficient implementations.

ricardoV94: I actually did that in my latest push. It short circuits if there's a dispatched implementation

ricardoV94:
Okay, this seems to be working reasonably well. Ready for review

tests/test_truncation.py
```diff
@@ -189,3 +193,203 @@ def test_censored_transform():
...
+    x = normal_op(loc, scale, name="x", size=100)
+
+    rng = aesara.shared(np.random.RandomState())
+    xt = truncate(x, lower=lower, upper=upper, rng=rng)
+    assert isinstance(xt.owner.op, TruncatedRV)
+    assert (xt.owner.inputs[0] if op_type == "icdf" else xt.owner.inputs[-1]) is rng
+    assert xt.type == x.type
+
+    # Check that original op can be used on its own
+    assert x.eval().shape
```
ricardoV94, line 261:
```suggestion
    assert x.eval().shape == (100,)
```

aeppl/truncation.py
```diff
@@ -129,3 +144,216 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+    logp = logp - lognorm
+
+    if is_lower_bounded:
+        logp = at.switch(value < lower_bound, -np.inf, logp)
+
+    if is_upper_bounded:
+        logp = at.switch(value <= upper_bound, logp, -np.inf)
+
+    if is_lower_bounded and is_upper_bounded:
+        logp = CheckParameterValue("lower_bound <= upper_bound")(
```
ricardoV94, line 344: Test this

aeppl/truncation.py
```diff
@@ -129,3 +144,216 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+    graph_inputs = graph_inputs[1:]
+    graph_inputs_ = graph_inputs_[1:]
+    *rv_inputs_, lower_, upper_ = (rng, *graph_inputs_)
+
+    rv_ = rv.owner.op.make_node(*rv_inputs_).default_output()
+
+    def loop_fn(truncated_rv, reject_draws, lower, upper, rng, *rv_inputs):
+        # We need to set default_update for scan to generate updates
+        next_rng, new_truncated_rv = rv.owner.op.make_node(rng, *rv_inputs).outputs
+        rng.default_update = next_rng
```
ricardoV94, line 254: TODO: Return updates manually instead of relying on default_update

aeppl/truncation.py
```diff
@@ -129,3 +144,216 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+        )
+        reject_draws = at.or_((truncated_rv < lower), (truncated_rv > upper))
+
+        return (truncated_rv, reject_draws), until(~at.any(reject_draws))
+
+    (truncated_rv_, reject_draws_), updates = scan(
+        loop_fn,
+        outputs_info=[
+            at.empty_like(rv_),
+            at.ones_like(rv_, dtype=bool),
```
ricardoV94, line 268: This can perhaps be simplified, now that scan does not struggle with default_updates as much

aeppl/truncation.py
brandonwillard: Yes, that's definitely needed before merging.

aeppl/truncation.py
```diff
@@ -129,3 +144,216 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+    if not (isinstance(rv.owner.op, RandomVariable) and rv.owner.op.ndim_supp == 0):
+        raise NotImplementedError(
+            f"truncation only implemented for univariate random variables, got {rv.owner.op}"
+        )
+
+    lower = at.as_tensor_variable(lower) if lower is not None else at.constant(-np.inf)
+    upper = at.as_tensor_variable(upper) if upper is not None else at.constant(np.inf)
+
+    if rng is None:
+        rng = shared(np.random.RandomState(), borrow=True)
```
brandonwillard, line 198: We need to use `RandomStream`s instead.

brandonwillard (changes requested):
This looks great; it only needs to be refactored to use `RandomStream`s and return updates.

aeppl/truncation.py
ricardoV94: In this case I think `RandomStream` causes more trouble than it helps. We can end up creating several intermediate RVs by the time we return the final graph, and `RandomStream` would be creating a new RNG shared variable and chaining them for each intermediate object, even if they are not returned. The only thing I really need is one RandomState variable

aeppl/truncation.py
```diff
@@ -129,3 +144,216 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+
+    # TODO: Scan does not return updates when a single step is performed, so this
+    #  will fail with max_n_steps = 1
+    truncated_rv = TruncatedRV(
+        base_rv_op=rv.owner.op,
+        inputs=graph_inputs_,
+        outputs=[tuple(updates.values())[0], truncated_rv_],
+        inline=True,
+    )(*graph_inputs)
+    truncated_rv.update = (truncated_rv.owner.inputs[-1], truncated_rv.owner.outputs[0])
```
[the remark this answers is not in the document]
ricardoV94: I was doing something similar to what the RandomStream does. AFAICT that `update` is not used anywhere internally, it was there for user convenience to access the update expression. But if we are returning that explicitly then there is no need for it.

aeppl/truncation.py
brandonwillard: > In this case I think `RandomStream` causes more trouble than it helps. We can end up creating several intermediate RVs by the time we return the final graph, and `RandomStream` would be creating a new RNG shared variable and chaining them for each intermediate object, even if they are not returned. The only thing I really need is one RandomState variable

`RandomStream` is the standard user interface for generating random samples in Aesara, so it **must** work with it.

If you believe there are issues with `RandomStream` itself, then those need to be addressed separately; otherwise, the things you mentioned are not issues that justify not using `RandomStream`.  

For instance, if you're concerned about the number of RNG objects created, then remember that `RandomStream` creates the shared RNG objects (i.e. they are not created in/by the compiled graph), so once `RandomStream.some_method` is called by the user, the work is already done.  If you're concerned that the user-level use of `RandomStream` is slow for this reason, then that would be surprising, because creating NumPy RNG objects should take no noticeable time.

The only in-graph concern is whether or not the `RandomVariable` `Op`s have been in-lined, but using `RandomStream` would make no difference in this case.

More importantly, you need to consider the **consistency** of the approach you're advocating.  Use of `RandomStream` guarantees that the sample results will be consistent with their user-level seeding, whereas the current approach does not.  

For instance, if the `rng` object in this code is used by two distinct but equivalent `RandomVariable`s and neither are in-lined, they will both produce the same sample values, and, if they are in-lined, they will produce distinct samples, and that isn't consistent.

There's more to it, but these reasons alone are enough to warrant the use of the `RandomStream` interface.

aeppl/truncation.py
```diff
@@ -122,3 +137,216 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+            at.ones_like(rv_, dtype=bool),
+        ],
+        non_sequences=[lower_, upper_, rng, *rv_inputs_],
+        n_steps=max_n_steps,
+        strict=True,
+    )
+
+    truncated_rv_ = truncated_rv_[-1]
+    convergence_ = ~at.any(reject_draws_[-1])
+    truncated_rv_ = Assert(
```
brandonwillard, line 272: This should probably be its own `CheckAndRaise` subclass that raises something like a `ValueError`.  The expectation is that `Assert`s can generally be ignored in performance-critical code, which is why we have an optional rewrite that removes them.  Since this is a numerical procedure that may or may not converge, it doesn't seem reasonable that one could remove `Assert`s for performance reasons and unknowingly end up using numerically incorrect results.

There are still some places in Aesara where `Assert` is used somewhat incorrectly in this way (e.g. broadcasting, I believe), but the expectation is that the underlying error will necessarily manifest in an `Op.perform` later down the line, so the `Assert` is just a emulating the place and type of error message that NumPy would return, but not avoiding the error altogether.  (Really, the issue here is that those `Assert`s should probably be the exact same exception types that NumPy would raise.)

ricardoV94: Yeah, I'll make a custom Error for this

aeppl/truncation.py
ricardoV94: I guess my issue with `RandomStream` is that it tries to do too much. It assumes what I want as updates and it introduces default_updates on the rng, which are a mess for any further graph manipulations and have been a constant source of bugs on my end.

aeppl/truncation.py
```diff
@@ -122,3 +138,237 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+    (truncated_rv_, reject_draws_), updates = scan(
+        loop_fn,
+        outputs_info=[
+            at.zeros_like(rv_),
+            at.ones_like(rv_, dtype=bool),
+        ],
+        non_sequences=[lower_, upper_, rng, size_, dtype_, *rv_inputs_],
+        n_steps=max_n_steps,
+        strict=True,
+    )
```
brandonwillard, line 290: Use `srng.gen` inside of `loop_fn`, then you can remove all of the `rng` arguments between `loop_fn` and `scan` and the updates return value in `loop_fn`.

aeppl/truncation.py
```diff
@@ -122,3 +138,237 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+        f"truncation did not converge in predefined {max_n_steps} steps"
+    )(truncated_rv_, convergence_)
+
+    truncated_rv = TruncatedRV(
+        base_rv_op=rv.owner.op,
+        inputs=graph_inputs_,
+        outputs=[tuple(updates.values())[0], truncated_rv_],
+        inline=True,
+    )(*graph_inputs)
+    updates = (truncated_rv.owner.inputs[-1], truncated_rv.owner.outputs[0])
```
brandonwillard, line 304:
```suggestion
    updates = {truncated_rv.owner.inputs[-1]: truncated_rv.owner.outputs[0]}
```

aeppl/truncation.py
```diff
@@ -122,3 +138,237 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+        ],
+        non_sequences=[lower_, upper_, rng, size_, dtype_, *rv_inputs_],
+        n_steps=max_n_steps,
+        strict=True,
+    )
+
+    truncated_rv_ = truncated_rv_[-1]
+    convergence_ = ~at.any(reject_draws_[-1])
+    truncated_rv_ = TruncationCheck(
+        f"truncation did not converge in predefined {max_n_steps} steps"
```
brandonwillard, line 295:
```suggestion
        f"Truncation did not converge in {max_n_steps} steps"
```

aeppl/truncation.py
```diff
@@ -122,3 +138,237 @@ def censor_logprob(op, values, base_rv, lower_bound, upper_bound, **kwargs):
...
+        truncated_rv, updates = _truncated(
+            rv.owner.op, lower, upper, srng, *rv.owner.inputs[1:]
+        )
+        return truncated_rv, updates
+    except NotImplementedError:
+        pass
+
+    # Variables with `_` suffix identify dummy inputs for the OpFromGraph
+    # We will use the Shared RNG variable directly because Scan demands it, even
+    # though it would not be necessary for the icdf OgFromGraph
```
brandonwillard, line 228:
```suggestion
```

brandonwillard (changes requested):

aeppl/truncation.py
ricardoV94: I did not do that because it would return a graph with two rngs, one for the `zeros_like` and one for the body inner loop.

brandonwillard: > I did not do that because it would return a graph with two rngs, one for the `zeros_like` and one for the body inner loop.

Wherever `RandomStream.gen` is used a new RNG object will be created.  There's nothing inherently bad about that, though; however, the current approach has all the same inconsistencies I mentioned earlier.

aeppl/truncation.py
brandonwillard: > I guess my issue with `RandomStream` is that it tries to do too much. It assumes what I want as updates and it introduces default_updates on the rng, which are a mess for any further graph manipulations and have been a constant source of bugs on my end.

Unless you faithfully reproduce what `RandomStream` does, the resulting graphs aren't guaranteed to be numerically consistent and/or they'll be considerably more difficult to optimize (e.g. due to the introduction of unnecessary subgraph dependencies); otherwise, if you reproduce what `RandomStream` already does, then it's just going to make the implementation more complicated/verbose.

ricardoV94: `RandomStream` does 3 things from what I gathered:
1. Create new SharedRandomStates from an initial seed
2. Add the updates of a new RV to a list and set them as default_update of the respective SharedRandomState
3. Reseed all RandomStates

My issue is only with step 2, and concretely with setting the `default_update`. I used to think it was necessary for use in Scan's but that does not seem to be the case. We just need to return updates explicitly instead of relying on the fallback behavior where it intercepts the `default_updates`.

The reason I don't like default_updates or hard-coded updates is that it complicates the manipulation of random graphs. I think we should be setting them prior to function compilation...

brandonwillard: > My issue is only with step 2. I used to think the `default_update` was necessary for use in Scan's but that does not seem to be the case. We just need to return updates explicitly instead of relying on the fallback behavior where it intercepts the `default_updates`.

As I stated above, the use of shared variables and `default_update` produces **independent** RNG state objects, and this is important for maintaining the applicability of optimizations.  Also, there currently isn't a numerically consistent non-shared variable/updates-based approach for updating an RNG state in `Scan` or `OpFromGraph`&mdash;unless `RandomVariable.inplace == True`, which isn't the case by default and breaks the underlying purity/reproducibility of our graphs.  

See https://github.com/aesara-devs/aesara/issues/738, https://github.com/aesara-devs/aesara/issues/898, and https://github.com/aesara-devs/aesara/issues/454 for more details about the aforementioned points.

> The reason I don't like default_updates or a frozen list of updates is that it complicates the manipulation of random graphs. Every time I want to work with a `Random Op` or an `OpFramGraph` that works with `SharedRandomStates`, I will have to worry about whether such variables came from a RandomStream or not. Even if I change the `default_update` (and this is making an assumption that the user did not put it there on purpose), there will now be an inconsistency between the `default_update` and the one stored in `RandomStream.updates()`.

Shared variables and `default_update` do not complicate anything that lies within the established framework for graph construction and manipulation; however, if you're trying to do something that **isn't** supported, then, yes, you'll need to account for things on your own.  `OpFromGraph` is a good example of something that **explicitly does not support updates** and, as a result, it does not work with `default_update`s or `Scan`s (that generate updates).  **This is an `OpFromGraph`-specific issue**, and it can only be solved by updating `OpFromGraph`.

Regardless, the alternative non-shared approaches I've seen are simply inconsistent, because they reuse the same RNG state(s) across random variables that are supposed to be entirely distinct, so we cannot use them.  Likewise, a user could add `default_update`s and shared variables of their own (i.e. outside of `RandomStream`), and **our code should still work**, so there's no avoiding the need to correctly handle those features.


> As a clear example, this aeppl helper is currently not safe to use with RVs with default updates, and was a source of quite bad bug in PyMC:
> 
> https://github.com/aesara-devs/aeppl/blob/87416830d7439cb64aba4d508361b14012227b2e/aeppl/utils.py#L68

That helper function (i.e. `change_rv_size`) was originally a quick hack to get something working; however, it was not a viable approach so we stopped using it (and it needs to be removed from this library).  The correct approach to that functionality is an `Op` that represents a change in the size of a `RandomVariable` (e.g. some sort of "replicate" `Op`).

Nevertheless, `change_rv_size` isn't used and shouldn't be, and not because of `RandomStream`, shared variables, or `default_update`, but because it produces a graph-level change in a `Variable` that occurs outside of the standard graph manipulation framework, and it has numerous fundamental issues due to **that** alone (e.g. any maps from/to the original variable will be broken once the results from that function are used in place of the original variable).

> The problem I have with RandomStream is that it makes it quite hard to manipulate graphs with RVs for the user convenience. I am fine with behaviors 1 and 3.

Again, it seems like the only place that actual problems appear are the places where updates are not immediately supported, which is currently **only** `OpFromGraph`.  Aside from that, I'm not aware of any other situations in which updates or `default_update` complicate graph manipulation.  

Regarding `Scan`, the answer is simple: return any updates explicitly created in a `Scan` loop function.  The only time that's really relevant is when a `Scan` loop function creates a `Scan`, and, in that scenario, the updates are already a return value that you're forced to deal with, so there are no surprises involved.  Again, this is something one must do regardless of `RandomVariable`s, `RandomStream`s, etc., because updates can be introduced in other ways.

Furthermore, now that https://github.com/aesara-devs/aesara/pull/910 is merged, the interaction between `Scan` and `default_update`s is more consistent.  If there are any other issues, we simply need to fix them like we did with that one, but we cannot introduce other inconsistencies or complicated code and graph structures in an attempt to avoid a narrow subset of problems that have already been resolved and/or have very straightforward solutions.

brandonwillard (dismissed):

rlouf:
Afair _truncation_ means that you _cannot_ observe any data and that you know that before observation. _Censoring_ is when we don't know the value of every data point, but only that it's greater or smaller than a given bound.

With that in mind it seems to me that the closest to truncation would be the following:

```python
import aesara.tensor as at

x_rv = srng.normal(0, 1)
y = at.clip(x_rv, -1, 1)
```

Where `y` is a random variable with a truncated distribution. To see this, it is enough to consider prior predictive sampling: `y` can only take values between -1 and 1. In the censored case, the prior predictive distribution should also return values that are out of bounds. I think the censored/truncated interface should be revamped as it currently doesn't reflect the mathematical meaning of these quantities.

ricardoV94:
Truncated means you cannot observe a value outside of a range, and you wouldn't know it it occurred. The generative graph involves rejection sampling.

Censoring mean you cannot observe a value outside of a range but you know when that occurs. The generative process involves clipping or rounding.

There is a good pymc example that explains the difference: and the Stan manual is also nice:

ricardoV94:
Your model generates censored data

rlouf:
No. To generate censored data your generate from the full distribution (`normal(0,1)`) and then apply a `>= threshold` filter to your prior samples to mimic the _observation_ process.

brandonwillard:
Are you guys essentially talking about how the endpoints of the `clip`ed intervals are/should be handled/understood?

rlouf:
> Are you guys essentially talking about how the endpoints of the `clip`ed intervals are/should be handled/understood?

Yes. I think I was wrong, it somehow didn't occur to me that   `at.clip` returns its endpoint.

brandonwillard:
> There's an implicit assumption about the meaning of the bounds when using `at.clip` that is not clearly stated, and not completely obvious. Saying that the value is equal to the value of the threshold is not the same as saying that it is greater than the threshold, these two correspond to different _types_. This probably touches on a deeper problem with the representation of these objects in the library.

Yes, there are some fundamental representation issues involved with using existing Aesara `Op`s for these types of derived measures.  Simply put, if samples from the original model graphs don't match the measures/densities we assign them, then we have a representation issue that needs to be fixed, and that's possibly the case with the representations of truncation and/or censoring used here.

ricardoV94:
> There's an implicit assumption about the meaning of the threshold values when using at.clip that is not clearly stated, and not completely obvious. Saying that the value is equal to the value of the threshold is not the same as saying that it is greater than the threshold, these two correspond to different types. This probably touches on broader design concerns wrt the representation of these objects in the library.

The logprob is correct (hopefully) for a data generative process that uses `at.clip`. This is somewhat analogous to a censoring process but it's not the only way of obtaining censored data and is also perhaps not the most intuitive.

Some other packages would use two vectors of data, one with the values and another with boolean/ categorical values that indicate whether a value was censored or not. You could try to handle this in aeppl as well.

Those have slightly different meanings at the edges. In the clip you assume edge values come from censoring for continuous distributions (ignoring precision issues), and a mixture process for discrete RVs (IIRC), whereas for the double vector approach you wouldn't have to assume or count as a mixture because it's stated by the user. The double vector retains more information about the data generation process.

Anyway, this PR was about truncation not censoring.

rlouf:
> Anyway, this PR was about truncation not censoring.

You're right, will open a separate discussion for my remaining questions about censoring.

One of my questions is still relevant for this PR: are you considering truncated random variables or truncated distributions here?

ricardoV94:
> > Anyway, this PR was about truncation not censoring.
> This is still relevant and important here:
> 
> > In this PR are we talking about truncated random variables or random variables with a truncated distribution? There isn't a 1:1 map between these two concepts. We also call truncated random variable a variable $Y$ that is defined such that $Y=X$ if $X < c$ and $Y=c$ if $X \geq c$ where $c$ a constant. This can be represented by `at.clip`, assuming in my example `x_rv` represents a random variable object (and not a distribution).
> 
> 

I think that still refers to censoring in the literature, not truncation. It's just one side censoring instead two side censoring. You can obtain those graphs with `y = at.clip(-np.inf, c)`. That's considered in the clip_logprob to generate a more efficient expression. There is also an issue for arbitrary censoring via `set_subtensor` here that you might be interested in:

rlouf:
Does this means this PR is about RVs with a truncated probabilistic distribution then?

ricardoV94:
> Does this means this PR is about RVs with a truncated probabilistic distribution then?

I guess so. I am not sure what is the difference between RVs and distributions that you were mentioning above. It's definitely about truncation, not censoring.

rlouf:
Ok, that's all I needed to know, and it's fine as long as it is clearly stated.  There is a difference between RVs with a truncated distribution (the sample space is the truncated interval), and a truncated RV which is the product of another RV  (who sample space can be the real line) with an indicator function.

When the terminology in the literature is so confused, here between measure-theory people and people only working with densities, it is important to be very specific about the relation between our representation and the different mathematical representation. This is not about splitting hair for the sake of it, but rather not ending up with the confusion and ambiguities that PPLs are usually ridden with.

rlouf:
As I expressed above, I have strong reservations when it comes to this PR and the meaning of the operator that is proposed. The problem can be seen clearly when comparing:

```python
import aesara.tensor as at

x = at.random.uniform(0, 10, name="x", size=100)
xt, _ = truncate(x, lower=5, upper=15, srng=srng)
```

to:

```python
import aesara.tensor as at

x = srng.uniform(0, 10)
xt = at.clip(x)
```

In the latter case, if we assume that $x$ represents a *random variable* the code can be correctly read as defining a new random variable $xt$ such that $xt = \operatorname{clip} \circ \operatorname{UniformRV}$. But if we admit this here, we also have to interpret the first snippet as defining a new random variable $xt$ such that $xt = \operatorname{truncate} \circ \operatorname{UniformRV}$. This is obviously not correct, and that is because $\operatorname{truncate}$ operates on *measures*.

I think that the abstractions that are present today in Aesara do not allow us to work on *measure* transformations since measures are not first-class citizens, but are present only implicitly.

Here is a modest proposal to alleviate these limitations. In this proposal we represent e.g. the measure with a standard normal density wrt the Lebesgue measure as 

```python
x_m = normal(0, 1)
```

the $\operatorname{truncate}$ operator acts on measures, so we can defined the truncated measure as

```python
x_t = truncate(normal(0, 1), lower=0)
```

to bridge the gap with the existing `RandomVariable` API in Aesara, we define the $\operatorname{sample}$ operator which takes a measure and returns an independent random variable with this measure each time it is called:

```python
x_rv = sample(rng, normal(0, 1))
```

here we retain the existing ambiguity there is in Aesara between random variables and elements of $E$ (which does not matter mathematically). The `RandomStream` interface can be retained with

```python
x_rv = srng.sample(normal(0, 1))
```

So the first code snippet above could be rewritten:

```python
import aesara.tensor as at
import aesara.tensor.random as ar

srng = ar.RandomStream(0)
x_m = truncate(ar.uniform(0, 10), lower=5, upper=15)
xt = srng.sample(x_m)
```

This interface has a few nice incidental properties:
- The `size` argument gets a new interpretation. Doing `x_rv = srng.sample(normal(0, 1), size=(10,))` means that we are defining 10 independent random variables with the same associated measure `normal(0, 1)` (we should probably rename it to `iid` or something similar);
- Provided we can make the `RandomStream` interface work this way, it would be much easier for one to add a custom measure and use it with this interface;
- I have a feeling this would make it easy to go the "explicitly representing rng key advancement" way if we ever choose to.

I can write an extended API proposal if you think this makes sense, and try to work out the potential challenges linked to this representation of probabilistic objects in Aesara and AePPL.

PR #144, opened 2022-04-29
Implement measurable transforms for division, subtraction and negation

ricardoV94:
Follow up to #26 

This implements measurable rewrites for the following type of graphs:

```python
x = at.random.gamma()
y = 1 / x
joint_logprob({y: y.type()})
```
As well as:
```
y = 5 / x
y = -x
y = 5 - x
```

It works by canonicalizing such operations to the form already understood by `find_measurable_transforms` (only a new condition for `Reciprocal`s was added)

```
(5 / x) -> 5 * reciprocal(x)
(-x) -> x * (-1)
5 - x -> 5 + (x * (-1))
```

These canonicalizations are only applied when MeasurableVariables are involved to minimize disruption of the underlying graph

### DIFF
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -10,7 +10,7 @@ from aesara.graph.features import AlreadyThere, Feature
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op
 from aesara.graph.rewriting.basic import GraphRewriter, in2out, node_rewriter
-from aesara.scalar import Add, Exp, Log, Mul, Reciprocal, TrueDiv
+from aesara.scalar import Add, Exp, Log, Mul, Neg, Reciprocal, Sub, TrueDiv
 from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.exceptions import NotScalarConstantError
 from aesara.tensor.rewriting.basic import (
@@ -293,6 +293,56 @@ def measurable_div_to_reciprocal_product(fgraph, node):
 
 
 @node_rewriter([Elemwise])
+def measurable_neg_to_product(fgraph, node):
+    """Convert negation of `MeasurableVariable`s to product with `-1`."""
+    if isinstance(node.op.scalar_op, Neg):
+        inp = node.inputs[0]
+        if not (inp.owner and isinstance(inp.owner.op, MeasurableVariable)):
+            return None
+
+        rv_map_feature: Optional[PreserveRVMappings] = getattr(
+            fgraph, "preserve_rv_mappings", None
+        )
+        if rv_map_feature is None:
+            return None  # pragma: no cover
+
+        # Only apply this rewrite if the variable is unvalued
+        if inp in rv_map_feature.rv_values:
+            return None  # pragma: no cover
+
+        return [at.mul(inp, -1.0)]
+
+
+@node_rewriter([Elemwise])
+def measurable_sub_to_neg(fgraph, node):
+    """Convert subtraction involving `MeasurableVariable`s to addition with neg"""
+    if isinstance(node.op.scalar_op, Sub):
+        measurable_vars = [
+            var
+            for var in node.inputs
+            if (var.owner and isinstance(var.owner.op, MeasurableVariable))
+        ]
+        if not measurable_vars:
+            return None  # pragma: no cover
+
+        rv_map_feature: Optional[PreserveRVMappings] = getattr(
+            fgraph, "preserve_rv_mappings", None
+        )
+        if rv_map_feature is None:
+            return None  # pragma: no cover
+
+        # Only apply this rewrite if there is one unvalued MeasurableVariable involved
+        if all(
+            measurable_var in rv_map_feature.rv_values
+            for measurable_var in measurable_vars
+        ):
+            return None  # pragma: no cover
+
+        minuend, subtrahend = node.inputs
+        return [at.add(minuend, at.neg(subtrahend))]
+
+
+@node_rewriter([Elemwise])
 def find_measurable_transforms(
     fgraph: FunctionGraph, node: Node
 ) -> Optional[List[Node]]:
@@ -383,11 +433,26 @@ def find_measurable_transforms(
 measurable_ir_rewrites_db.register(
     "measurable_div_to_reciprocal_product",
     measurable_div_to_reciprocal_product,
-    -1,
+    -5,
     "basic",
     "transform",
 )
 
+measurable_ir_rewrites_db.register(
+    "measurable_neg_to_product",
+    measurable_neg_to_product,
+    -5,
+    "basic",
+    "transform",
+)
+
+measurable_ir_rewrites_db.register(
+    "measurable_sub_to_neg",
+    measurable_sub_to_neg,
+    -5,
+    "basic",
+    "transform",
+)
 
 measurable_ir_rewrites_db.register(
     "find_measurable_transforms",
--- a/aeppl/transforms.py
+++ b/aeppl/transforms.py
@@ -10,8 +10,9 @@ from aesara.graph.features import AlreadyThere, Feature
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op
 from aesara.graph.rewriting.basic import GraphRewriter, in2out, node_rewriter
-from aesara.scalar import Add, Exp, Log, Mul
+from aesara.scalar import Add, Exp, Log, Mul, Reciprocal, TrueDiv
 from aesara.tensor.elemwise import Elemwise
+from aesara.tensor.exceptions import NotScalarConstantError
 from aesara.tensor.rewriting.basic import (
     register_specialize,
     register_stabilize,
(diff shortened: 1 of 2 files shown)


brandonwillard (commented):
Can you add a little more context to the description?

ricardoV94:
> Can you add a little more context to the description?

Done

aeppl/opt.py
```diff
@@ -292,6 +294,20 @@ def naive_bcast_rv_lift(fgraph, node):
...
+    """Convert 5/3 to 5 * 1/3"""
+    if isinstance(node.op.scalar_op, TrueDiv):
+        numerator, denominator = node.inputs
+        # Check if numerator is 1
+        try:
+            if at.get_scalar_constant_value(numerator) == 1:
+                return [at.reciprocal(denominator)]
+        except NotScalarConstantError:
+            pass
+        return [at.mul(numerator, at.reciprocal(denominator))]
```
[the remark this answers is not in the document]
brandonwillard: This looks like something that's better done with unification at the moment.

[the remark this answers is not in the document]
ricardoV94: Why do you say so?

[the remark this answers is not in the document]
brandonwillard: The tensor algebra canonicalizations shouldn't (and don't need to) be changed to suit a single purpose like this.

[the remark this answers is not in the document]
ricardoV94: I made the rewrite(s) specific only to terms with MeasurableVariables. This should be no more disruptive than some of the other rewrites implemented in `opt.py`. Does this sound like a more sound approach?

ricardoV94:
Added other rewrites that cover negation and subtraction of MeasurableVariables

aeppl/transforms.py
```diff
@@ -255,13 +256,99 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
     return input_logprob + jacobian
 
 
+@local_optimizer([Elemwise])
+def measurable_div_to_reciprocal_product(fgraph, node):
+    """Convert divisions involving `MeasurableVariable`s to product with reciprocal."""
```
brandonwillard, line 261:
```suggestion
    r"""Convert divisions involving `MeasurableVariable`\s to products with their reciprocals."""
```

aeppl/transforms.py
```diff
@@ -255,13 +256,99 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
...
+            if at.get_scalar_constant_value(numerator) == 1:
+                return [at.reciprocal(denominator)]
+        except NotScalarConstantError:
+            pass
+        return [at.mul(numerator, at.reciprocal(denominator))]
+
+
+@local_optimizer([Elemwise])
+def measurable_neg_to_product(fgraph, node):
+    """Convert negation of `MeasurableVariable`s to product with `-1`."""
```
brandonwillard, line 297:
```suggestion
   r"""Convert negations of `MeasurableVariable`\s to products with `-1`."""
```

aeppl/transforms.py
```diff
@@ -255,13 +256,99 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
...
+        # Only apply this rewrite if the variable is unvalued
+        if inp in rv_map_feature.rv_values:
+            return None
+
+        return [at.mul(inp, -1.0)]
+
+
+@local_optimizer([Elemwise])
+def measurable_sub_to_neg(fgraph, node):
+    """Convert subtraction involving `MeasurableVariable`s to addition with neg"""
```
brandonwillard, line 318:
```suggestion
    r"""Convert subtraction involving `MeasurableVariable`\s to additions with `neg`."""
```

aeppl/transforms.py
```diff
@@ -341,6 +430,30 @@ def find_measurable_transforms(
...
+)
+
+measurable_ir_rewrites_db.register(
+    "measurable_sub_to_neg",
+    measurable_sub_to_neg,
+    -5,
+    "basic",
+    "transform",
+)
+
```
brandonwillard, line 456: These are rewrites that appear to apply to all `MeasurableVariable`s regardless of transforms.  They're actually adjustments to the canonical form of `MeasurableVariable`s.  As such, they should at least be somewhere more general&mdash;like `aeppl.tensor`.

brandonwillard (changes requested):

aeppl/transforms.py
ricardoV94: I initially had them in `opt.py` actually. What do you think makes more sense, there or in `tensor.py`?

brandonwillard: Since the rewrites deal specifically with algebraic tensor operations, `aeppl.tensor` is probably the most suitable in name.

brandonwillard (changes requested):
I'm very unsure about these fundamental (implicit) canonicalization changes made to rational expressions.  Those could really create issues down the road, and they're something that should probably be used in Aesara as well, or not at all.  In other words, if they're justified here, why aren't they justified for general use in Aesara?

Anyway, we might be able to try going forward with this&mdash;after renaming that class&mdash;but I still think it needs more consideration, especially since adding effective canonicalizations isn't usually necessary (i.e. functionality-specific local rewrites can make the relevant rational form considerations instead).

ricardoV94:
> In other words, if they're justified here, why aren't they justified for general use in Aesara?

I think it makes sense to canonicalize in Aesara, except if weird float precision issues may crop up in general. We already have an issue opened for that, so if it goes through we can revert these ones.

But here there's also a question of context. We know that in Aeppl most cases of unvalued RVs are meant to be replaced. If for some reason these canonicalizations are not desired, users can remove the rewrite from the standard database.

We also have other canonicalizations that don't translate to measurable variables directly like the broadcast one.

ricardoV94:
Alternatively, if we convert all unvalued RVs to MeasurableVariables before applying these canonicalizations, they will be undone when they are not needed by the same machinery in here: https://github.com/aesara-devs/aeppl/blob/394dc72a536f137fbeded74a196769baba807dc9/aeppl/opt.py#L321

aeppl/transforms.py
```diff
@@ -413,6 +415,19 @@ def log_jac_det(self, value, *inputs):
...
+    name = "inverse"
+
+    def forward(self, value, *inputs):
+        return at.inv(value)
+
+    def backward(self, value, *inputs):
+        return at.inv(value)
+
+    def log_jac_det(self, value, *inputs):
+        return -at.log(value**2)
```
ricardoV94, line 428:
```suggestion
        return -2 *at.log(value)
```

I wonder if this is something we could have as a specialization rewrite or is one of those that is not universally better

ricardoV94:
@brandonwillard I have not addressed the concerns about intermediate rewrites that simply canonicalize intermediate variables, besides restricting those rewrites to "unvalued measurable variables". 

IMO until we figure out what we want to do with unmeasured RandomVariables in our graphs (as discussed in #174 and #85) we could keep assuming they are there to be replaced anyway, in which case these rewrites don't do any damage and make it simpler to cover a wider range of measurable Elemwise operations.

Anyway, I re-requested your review to check that this is the only concern left and discuss it.

brandonwillard:
> 

We can always make these rewrites elective (i.e. a user must explicitly load/use them) for the time being; otherwise, adding rewrites that change the expectations of other rewrites (i.e. affect and/or determine canonical/normal forms) leads to serious downstream problems.  For instance, if `a / b` is always converted to `a * (1 / b)` then subsequent rewrites expecting `a / b` (sub-)expressions are rendered useless.  This could manifest as an inability to use certain basic Aesara rewrites with AePPL IR, or, at the very least, it could produce AePPL IR graphs that are inconsistent with the ones produced by Aesara for equivalent graphs.

ricardoV94:
Thanks for the reply.

I am not totally convinced why this matters here. Assuming we introduce other rewrites in the future that would try to match `a = neg(x)`, I assume they would also want to match `b = mul(-1, x)`. Therefore such rewrite would need to 1) try to match both, in which case it doesn't matter that we changed from `a` to `b` or 2) save work and only match one, in which case it helps to have an "Aeppl canonicalized form" (`b`). 

I understand why in Aesara we might not want to do that because of numerical stability, but here we are not dealing with direct computations yet, and we don't even know which, if any, IR form will lead to a more stable logprob (and the user much less), specially because this may be used as a building block of arbitrarily complex valued-variables.

Note that the outputs of these intermediate rewrites will always be picked up by `find_measurable_transforms` and converted to a MeasurableVariable, so the final outcome would be completely equivalent if I instead expanded the `find_measurable_transforms` rewrite to be able to match different variations and skip the intermediate forms. 

The only side-effect happens with unvalued-orphan `RandomVariable`s which we haven't decided on a policy for yet.

brandonwillard:
In general, it looks like the transform approach in this PR is rather buggy and acceptable changes to the rewrite orders cause numerous problems.  After fixing something like #170 and running this becomes **very** apparent.

I'm going to close this and open a replacement PR that fixes this, and the `MeasurableElemwise` approach as well.

Issue #148, opened 2022-06-21
`test_scan.py::test_initial_values` failing with latest Aesara release


ricardoV94:
This showed up in #147 (https://github.com/aesara-devs/aeppl/runs/6918324742?check_suite_focus=true)

https://github.com/aesara-devs/aeppl/blob/cc78f30b5ed89e5b247b88d697467a76cc1e424e/tests/test_scan.py#L349

ricardoV94:
@brandonwillard suggestion for what might be going on:

> it has something to do with test values assigned to NominalVariables
> remember that those kinds of Variables are identical when their IDs are equal
> anyway, the Scan manipulations in aeppl needed some kind of adjustment
> to account for NominalVariables appropriately
> and their associated test values
> it might be that the Scan inner-graphs aren't creating NominalVariables that are properly ordered
> e.g. if an inner-graph is constructed from other graphs that have the same NominalVariables in them, then the new inner-graph needs to reorder/recreate the NominalVariables so that they're deduplicated and represent all the new inner-graph inputs distinctly

PR #151, opened 2022-06-30
Fix check for missing rv in factorized_joint_logprob

larryshamalama:
Currently, random variables that are not provided with their value variable equivalent can result in a `KeyError` in `factorized_joint_logprob` when given `warn_missing_rvs=False`. This error rose when working with a graph corresponding to a Gaussian random walk.

```python
import aesara.tensor as at

mu = at.random.normal(loc=2, scale=0.1)
sigma = at.random.gamma(shape=1, rate=1)
innovation_dist = at.random.normal(mu[..., None], sigma[..., None], size=(10,))
rv_out = at.cumsum(innovation_dist, axis=-1) # gaussian random walk
```

The following code chunk yields a `KeyError`:

```python
factorized_joint_logprob(
    {
        rv_out: rv_out.copy(),
    }, 
    extra_rewrites=TransformValuesOpt({}),
    use_jacobian=True,
    warn_missing_rvs=False
)
```

but setting `warn_missing_rvs=True` only issues warnings.

In this PR, so far, I fix the case where users decide to not issue warnings.

### DIFF
--- a/tests/test_scan.py
+++ b/tests/test_scan.py
@@ -346,6 +346,7 @@ def test_scan_joint_logprob(require_inner_rewrites):
 
 
 @aesara.config.change_flags(compute_test_value="raise")
+@pytest.mark.xfail(reason="see #148")
 def test_initial_values():
     srng = at.random.RandomStream(seed=2320)
 
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -122,11 +122,12 @@ def factorized_joint_logprob(
         if not outputs:
             continue
 
-        if warn_missing_rvs and any(o not in updated_rv_values for o in outputs):
-            warnings.warn(
-                "Found a random variable that was neither among the observations "
-                f"nor the conditioned variables: {node.outputs}"
-            )
+        if any(o not in updated_rv_values for o in outputs):
+            if warn_missing_rvs:
+                warnings.warn(
+                    "Found a random variable that was neither among the observations "
+                    f"nor the conditioned variables: {node.outputs}"
+                )
             continue
 
         q_value_vars = [replacements[q_rv_var] for q_rv_var in outputs]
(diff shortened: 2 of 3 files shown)


ricardoV94:
Can you add a test? Your example should be fine just assert that the warning raised in one case but not the other, but the output is the same.

larryshamalama:
Done! There was already a test for this, I just added a verification for `warn_missing_rvs=False`

ricardoV94 (approved):

larryshamalama:
Failing test as described in #148:

https://github.com/aesara-devs/aeppl/blob/cc78f30b5ed89e5b247b88d697467a76cc1e424e/tests/test_scan.py#L349

ricardoV94:
You can just cherry pick this commit: https://github.com/aesara-devs/aeppl/commit/e0f9623fa6734061e05850eb14c72e18ddd7027f

ricardoV94:
You should have the two commits separate.

One for the xfail and the other for the bugfix and new test condition.

larryshamalama:
> You should have the two commits separate.
> 
> One for the xfail and the other for the bugfix and new test condition.

Done! Thanks for the clarification

ricardoV94:
Small nitpick the test and bugfix make sense to go in the same commit

larryshamalama:
After several attempts and many force pushes, I think that I got it right (?) 😅

tests/test_scan.py
```diff
@@ -345,7 +345,7 @@ def scan_fn(mus_t, sigma_t, Y_t_val, S_t_val, Gamma_t):
     assert np.allclose(y_logp_val, y_logp_ref_val)
 
 
-@aesara.config.change_flags(compute_test_value="raise")
```
ricardoV94, line 348: Ah... we should not delete this previous decorator. The xfail still works with it no?

larryshamalama: Oops... I made the edits too hastily. I corrected this now

ricardoV94 (approved):

PR #154, opened 2022-07-05
Introduce graph rewrite for mixture sub-graphs defined via `Switch`

### LINKED ISSUE #77: Support mixtures defined with `switch`
We need to add support for mixtures defined using the `Elemwise` +  `Switch` `Op`.

The following is a simple example that isn't supported:
```python
import aesara.tensor as at

from aeppl.joint_logprob import factorized_joint_logprob


srng = at.random.RandomStream(seed=2320)

I_rv = srng.bernoulli(0.5, size=10, name="I")
Y_rv = srng.gamma(0.5, 0.5, size=10, name="Y")

Z_rv = at.switch(I_rv, X_rv, Y_rv)
Z_rv.name = "Z"

z_vv = Z_rv.clone()
i_vv = I_rv.clone()
logp_parts = factorized_joint_logprob({Z_rv: z_vv, I_rv: i_vv})
```

larryshamalama:
Closes #77.

Currently, I am opening this PR to ask for pointers. So far, many lines are copy pasted from `mixture_replace` in the same file. Hopefully, this attempt is on the right track, although it is not working at the moment. The output of a mixture defined by `switch` is showcased in [this gist](https://gist.github.com/larryshamalama/b2e35ec250d6de40974235ed0ee7f53c)

### DIFF
--- a/aeppl/mixture.py
+++ b/aeppl/mixture.py
@@ -11,7 +11,9 @@ from aesara.graph.rewriting.basic import (
     pre_greedy_node_rewriter,
 )
 from aesara.ifelse import ifelse
+from aesara.scalar.basic import Switch
 from aesara.tensor.basic import Join, MakeVector
+from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.random.rewriting import (
     local_dimshuffle_rv_lift,
     local_subtensor_rv_lift,
@@ -19,6 +21,7 @@ from aesara.tensor.random.rewriting import (
 from aesara.tensor.shape import shape_tuple
 from aesara.tensor.subtensor import (
     as_index_literal,
+    as_nontensor_scalar,
     get_canonical_form_slice,
     is_basic_idx,
 )
@@ -251,7 +254,6 @@ def mixture_replace(fgraph, node):
     From these terms, new terms ``Z_rv[i] = mixture_comps[i][i == I_rv]`` are
     created for each ``i`` in ``enumerate(mixture_comps)``.
     """
-
     rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
 
     if rv_map_feature is None:
@@ -292,6 +294,56 @@ def mixture_replace(fgraph, node):
     if aesara.config.compute_test_value != "off":
         # We can't use `MixtureRV` to compute a test value; instead, we'll use
         # the original node's test value.
+        if not hasattr(old_mixture_rv.tag, "test_value"):
+            compute_test_value(node)
+
+        new_mixture_rv.tag.test_value = old_mixture_rv.tag.test_value
+
+    if old_mixture_rv.name:
+        new_mixture_rv.name = f"{old_mixture_rv.name}-mixture"
+
+    return [new_mixture_rv]
+
+
+@node_rewriter((Elemwise,))
+def switch_mixture_replace(fgraph, node):
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    if not isinstance(node.op.scalar_op, Switch):
+        return None  # pragma: no cover
+
+    old_mixture_rv = node.default_output()
+    # idx, component_1, component_2 = node.inputs
+
+    mixture_rvs = []
+
+    for component_rv in node.inputs[1:]:
+        if not (
+            component_rv.owner
+            and isinstance(component_rv.owner.op, MeasurableVariable)
+            and component_rv not in rv_map_feature.rv_values
+        ):
+            return None
+        new_node = assign_custom_measurable_outputs(component_rv.owner)
+        out_idx = component_rv.owner.outputs.index(component_rv)
+        new_comp_rv = new_node.outputs[out_idx]
+        mixture_rvs.append(new_comp_rv)
+
+    mix_op = MixtureRV(
+        2,
+        old_mixture_rv.dtype,
+        old_mixture_rv.broadcastable,
+    )
+    new_node = mix_op.make_node(
+        *([NoneConst, as_nontensor_scalar(node.inputs[0])] + mixture_rvs)
+    )
+
+    new_mixture_rv = new_node.default_output()
+
+    if aesara.config.compute_test_value != "off":
         if not hasattr(old_mixture_rv.tag, "test_value"):
             compute_test_value(node)
 
@@ -368,7 +420,8 @@ def logprob_MixtureRV(
 logprob_rewrites_db.register(
     "mixture_replace",
     EquilibriumGraphRewriter(
-        [mixture_replace], max_use_ratio=aesara.config.optdb__max_use_ratio
+        [mixture_replace, switch_mixture_replace],
+        max_use_ratio=aesara.config.optdb__max_use_ratio,
     ),
     0,
     "basic",
--- a/tests/test_mixture.py
+++ b/tests/test_mixture.py
@@ -3,13 +3,14 @@ import aesara.tensor as at
 import numpy as np
 import pytest
 import scipy.stats.distributions as sp
-from aesara.graph.basic import Variable
+from aesara.graph.basic import Variable, equal_computations
 from aesara.tensor.random.basic import CategoricalRV
 from aesara.tensor.shape import shape_tuple
 from aesara.tensor.subtensor import as_index_constant
 
 from aeppl.joint_logprob import factorized_joint_logprob, joint_logprob
-from aeppl.mixture import expand_indices
+from aeppl.mixture import MixtureRV, expand_indices
+from aeppl.rewriting import construct_ir_fgraph
 from tests.test_logprob import scipy_logprob
 from tests.utils import assert_no_rvs
 
@@ -67,7 +68,14 @@ def test_mixture_basics():
 
 
 @aesara.config.change_flags(compute_test_value="warn")
-def test_compute_test_value():
+@pytest.mark.parametrize(
+    "op_constructor",
+    [
+        lambda _I, _X, _Y: at.stack([_X, _Y])[_I],
+        lambda _I, _X, _Y: at.switch(_I, _X, _Y),
+    ],
+)
+def test_compute_test_value(op_constructor):
 
     srng = at.random.RandomStream(29833)
 
@@ -82,7 +90,7 @@ def test_compute_test_value():
     i_vv = I_rv.clone()
     i_vv.name = "i"
 
-    M_rv = at.stack([X_rv, Y_rv])[I_rv]
+    M_rv = op_constructor(I_rv, X_rv, Y_rv)
     M_rv.name = "M"
 
     m_vv = M_rv.clone()
@@ -705,3 +713,49 @@ def test_mixture_with_DiracDelta():
     logp_res = factorized_joint_logprob({M_rv: m_vv, I_rv: i_vv})
 
     assert m_vv in logp_res
+
+
+def test_switch_mixture():
+    srng = at.random.RandomStream(29833)
+
+    X_rv = srng.normal(-10.0, 0.1, name="X")
+    Y_rv = srng.normal(10.0, 0.1, name="Y")
+
+    I_rv = srng.bernoulli(0.5, name="I")
+    i_vv = I_rv.clone()
+    i_vv.name = "i"
+
+    Z1_rv = at.switch(I_rv, X_rv, Y_rv)
+    z_vv = Z1_rv.clone()
+    z_vv.name = "z1"
+
+    fgraph, _, _ = construct_ir_fgraph({Z1_rv: z_vv, I_rv: i_vv})
+
+    assert isinstance(fgraph.outputs[0].owner.op, MixtureRV)
+    assert not hasattr(
+        fgraph.outputs[0].tag, "test_value"
+    )  # aesara.config.compute_test_value == "off"
+    assert fgraph.outputs[0].name is None
+
+    Z1_rv.name = "Z1"
+
+    fgraph, _, _ = construct_ir_fgraph({Z1_rv: z_vv, I_rv: i_vv})
+
+    assert fgraph.outputs[0].name == "Z1-mixture"
+
+    # building the identical graph but with a stack to check that mixture computations are identical
+
+    Z2_rv = at.stack((X_rv, Y_rv))[I_rv]
+
+    fgraph2, _, _ = construct_ir_fgraph({Z2_rv: z_vv, I_rv: i_vv})
+
+    assert equal_computations(fgraph.outputs, fgraph2.outputs)
+
+    z1_logp = joint_logprob({Z1_rv: z_vv, I_rv: i_vv})
+    z2_logp = joint_logprob({Z2_rv: z_vv, I_rv: i_vv})
+
+    # below should follow immediately from the equal_computations assertion above
+    assert equal_computations([z1_logp], [z2_logp])
+
+    np.testing.assert_almost_equal(0.69049938, z1_logp.eval({z_vv: -10, i_vv: 0}))
+    np.testing.assert_almost_equal(0.69049938, z2_logp.eval({z_vv: -10, i_vv: 0}))


rlouf:
I think the code in the notebook is short enough to include it directly in your comment (it's easier to follow when possible). Besides that do you have a specific question/problem or is this just work in progress?

larryshamalama:
> I think the code in the notebook is short enough to include it directly in your comment (it's easier to follow when possible). Besides that do you have a specific question/problem or is this just work in progress?

Both! I'm writing them as a code review, I'm just a bit slow

aeppl/mixture.py
```diff
@@ -300,6 +302,67 @@ def mixture_replace(fgraph, node):
...
+        new_comp_rv = new_node.outputs[out_idx]
+        mixture_rvs.append(new_comp_rv)
+
+    # Replace this sub-graph with a MixtureRV
+    mix_op = MixtureRV(
+        2,
+        old_mixture_rv.dtype,
+        old_mixture_rv.broadcastable,
+    )
+    new_node = mix_op.make_node(*([NoneConst, mixing_indices] + mixture_rvs))
```
larryshamalama, line 348: I believe that the lines here at the most important.

- 2 as first argument to `MixtureRV` because I'm assuming that `mixing_indices` is just a scalar. I imagine that this will be extended to `1 + len(mixing_indices)` as in `mixture_replace`;
- I put `NoneConst` as the first input to `mix_op.make_node` as I am not sure what axis does `at.switch` work with.

aeppl/mixture.py
```diff
@@ -300,6 +302,67 @@ def mixture_replace(fgraph, node):
...
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    old_mixture_rv = node.default_output()
+
+    if len(node.inputs) != 3:
+        return None  # graph structure should be [index, component1, component2]
+
+    mixing_indices = node.inputs[0]
+    mixture_res = node.inputs[1:]
```
larryshamalama, line 327: I believe that `switch` statements such as `at.switch(I_rv, X_rv, Y_rv)` **always** takes three arguments: the index, component 1 and component 2. Is this true? I currently define `mixture_res` as the `[X_rv, Y_rv]`.

aeppl/mixture.py
```diff
@@ -371,3 +434,14 @@ def logprob_MixtureRV(
...
+
+logprob_rewrites_db.register(
+    "switch_mixture_replace",
+    EquilibriumOptimizer(
+        [switch_mixture_replace], max_use_ratio=aesara.config.optdb__max_use_ratio
+    ),
+    0,
+    "basic",
+    "mixture",
+)
```
larryshamalama, line 447: These lines follow the registration of `mixture_replace` in `logprob_rewrites_db` very closely.

larryshamalama (commented):
The toy example that I'm currently working with is the one in the gist linked above, which is:

```python
srng = at.random.RandomStream(seed=2320)

I_rv = srng.bernoulli(0.5, name="I")
X_rv = srng.normal(loc=-5, scale=0.1, name="X")
Y_rv = srng.normal(loc=5, scale=0.1, name="Y")

Z_rv = at.switch(I_rv, X_rv, Y_rv)
Z_rv.name = "Z"

z_vv = Z_rv.clone()
i_vv = I_rv.clone()
logp = joint_logprob({Z_rv: z_vv, I_rv: i_vv})
```

Currently, providing numerical values for `z_vv` and `i_vv` via `logp.eval({z_vv: 5, i_vv: 1})` yields (what looks like) correct log-probability quantities:

```python
logp.eval({z_vv: 5, i_vv: 0}) # array(-4999.30950062)
logp.eval({z_vv: 5, i_vv: 1}) # array(0.69049938)
```

However, `logp.eval()` yields a `MissingInputError`. My current understanding is twofold:

1- the subgraph substitution is _happening_ due to the obtention of log-probabilities with numerical values provided, but this process needs to be polished;
2- `logp.eval()` should yield the log-probability with `z_vv` and `i_vv` evaluated at random quantities, which it is currently not doing. Should this be the expected behaviour? I am not sure about this.

Happy to hear any pointers or suggestions about this progress

aeppl/mixture.py
```diff
@@ -300,6 +302,67 @@ def mixture_replace(fgraph, node):
...
+        # We create custom types for the mixture components and assign them
+        # null get_measurable_outputs dispatches so that they aren't
+        # erroneously encountered in places like factorized_joint_logprob.
+        new_node = assign_custom_measurable_outputs(component_rv.owner)
+        out_idx = component_rv.owner.outputs.index(component_rv)
+        new_comp_rv = new_node.outputs[out_idx]
+        mixture_rvs.append(new_comp_rv)
+
+    # Replace this sub-graph with a MixtureRV
+    mix_op = MixtureRV(
```
ricardoV94, line 343: If you are just reusing the MixtureRV class and logprob you could just make this rewrite return the `stack((rv1, rv2))` form that the existing mixture machinery recognizes.

larryshamalama: Oh, I didn't think of this. However, would stacking add some computational burden? I.e. `switch` -> `stack` -> `MixtureRV` vs. `switch` -> `MixtureRV`

ricardoV94: That extra step should be nearly irrelevant. 

More important is whether we want to define the logprob in terms of switch instead of indexing. That will require some investigation and benchmarking. For now, since the end logprob function is the same, it doesn't matter.

larryshamalama: Got it 👍  I'll give it a shot

aeppl/mixture.py
```diff
@@ -300,6 +302,67 @@ def mixture_replace(fgraph, node):
...
+    if len(node.inputs) != 3:
+        return None  # graph structure should be [index, component1, component2]
+
+    mixing_indices = node.inputs[0]
+    mixture_res = node.inputs[1:]
+
+    # We loop through mixture components and collect all the array elements
+    # that belong to each one (by way of their indices).
+    mixture_rvs = []
+    for i, component_rv in enumerate(mixture_res):
```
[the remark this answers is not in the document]
larryshamalama: Thanks for the comment. I am leaving this "unresolved" for now and to be addressed later once I figure out the structure of how to rewrite nested `switch`es into a `stack`.

aeppl/mixture.py
```diff
@@ -300,6 +302,67 @@ def mixture_replace(fgraph, node):
...
+        # the original node's test value.
+        if not hasattr(old_mixture_rv.tag, "test_value"):
+            compute_test_value(node)
+
+        new_mixture_rv.tag.test_value = old_mixture_rv.tag.test_value
+
+    if old_mixture_rv.name:
+        new_mixture_rv.name = f"{old_mixture_rv.name}-mixture"
+
+    return [new_mixture_rv]
```
larryshamalama, line 363: Thinking sort of out loud here. I have made progress on how to rewrite nested switches into a stack and I'm at the point where I should test my logic on more graphs (starting in NumPy).

However, I believe that I should make use of `mixture_replace` here. Should I be returning `mixture_replace(fgraph, node)` with a newly constructed `fgraph`? If so, I believe that the pseudo-code would look something like this:

```python
fgraph = FunctionGraph(...)
new_node = at.stack([component1, component2, ...])

return mixture_replace(fgraph, new_node)
```

Does this approach seem reasonable?

larryshamalama (commented):
Question on how to make use of `mixture_replace` to utilize the graph rewriting functionality for mixtures built using `stack`s

aeppl/mixture.py
ricardoV94: Your rewrite should not have to construct a new fgraph (it takes it as input). I guess you can call `mixture_replace` directly, but if you return just the `new_node`, the `mixture_replace` will later be called on that node automatically anyway. @brandonwillard any preference here for directly calling `mixture_replace`?

brandonwillard: If we believe there will be no other rewrite that could/should be applied to the intermediate results, then, sure, calling `mixture_replace` directly is fine.  It might mess with some profiling results, but nothing serious.

larryshamalama:
I spent the time since our last meeting thinking about this problem from a mathematical perspective (and other PyMC-related things). Most of my progress is not yet in code.

## Problem

I believe that graphs of nested `switch`es can be thought as binary trees where each node has 0 or 2 children. In the case of 0 children, that is if the node is a component and, in the case of 2 children, the node is another `switch` statement. To my understanding, the problem becomes the following: given a binary tree with $k$ `switch`es, $k$ index variables $I_1, \dots, I_k$ and $k + 1$ leaves $A_1, \dots, A_{k+1}$, can we come up with a general function $h: \\{0, 1\\}^k \rightarrow \\{0, \dots, k \\}$ such that $h(I_1, \dots, I_k)$ yields the correct node $A_j$ as in the binary tree defined by nested `switch`es?

## Examples

Here are three basic examples that I've written by hand. All of these are graphs of depth 2, but it's a start. Let `srng = at.random.RandomStream(seed=2320)` in all examples below.

### Example 1

```python
I1_rv = srng.bernoulli(0.5, name="I1")
A_rv = srng.normal(-5, 0.1, name="A")
B_rv = srng.normal(5, 0.1, name="B")

C_rv = at.switch(I1_rv, A_rv, B_rv)
C_rv.name = "C"

I2_rv = srng.bernoulli(0.5, name="I2")
D_rv = srng.normal(10, 0.1, name="D")

Z_rv = at.switch(I2_rv, C_rv, D_rv)
Z_rv.name = "Z"
```

*Below, $I_2 = 1$ label missing for C/A edge and $I_2 = 0$ label missing for C/B edge

image

Here, $h(I_1, I_2) = I_1 + I_1 I_2$.

### Example 2

image

Here, $h(I_1, I_2) = 2I_1 + I_2(1 - I_1)$.

### Example 3

brandonwillard:
> I believe that graphs of nested `switch`es can be thought as binary trees where each node has 0 or 2 children.

Yes, your high-level scalar binary arithmetic characterization looks like it's on track, but don't forget that `switch`es take arbitrary tensors as arguments, which means that you'll need to extend your characterization to&mdash;at least&mdash;vector spaces.

Regardless, the things that need to be addressed for an implementation often have more to do with the graph traversal and reuse of existing logic.

For instance, it might not be reasonable to walk a graph and attempt to identify nested `switch` subgraphs and directly implement the binary arithmetic you've outlined here.  The end result could easily involve a lot of custom logic and make this feature very difficult to maintain (e.g. see `Scan` in Aesara).

Ideally, a "local" rewrite of the relevant `switch`es would suffice, and, when those rewrites are applied repeatedly, they would cover all nested `switch`-containing graph cases as well.

This is why the indexing/`Subtensor`-based approach is so appealing.  If we can convert `switch`es into equivalent `Subtensor` graphs (e.g. `switch(a, b, c)` converted to `stack([b, c])[a_new]`), then we can reuse all the existing `Subtensor`-based mixture code and address the nested `switch` cases via `stack`s (i.e. `Join` or `MakeVector` `Op`s) and `Subtensor` rewrites&mdash;some of which may already exist.  This approach has the added benefit that any missing `stack`/`Subtensor` rewrites we implement for these purposes will then be available for more general use, which would also expand the capabilities of Aesara more broadly.

brandonwillard:
Actually, if I'm understanding this correctly, you could also be attempting to describe the logic for nested _mixtures_, which might be helpful for rewrites that target our `MixtureRV` IR more directly.  

More specifically, once we identify `stack([b, c])[a_new]` as a mixture, it's turned into a single `MixtureRV` node, so nested `switch`es would ultimately look like `mixture(a_new, mixture(a_2, b_2, c_2), mixture(a_3, b_3, c_3))`, and those could be rewritten much more directly, but the question is "How?"

aeppl/mixture.py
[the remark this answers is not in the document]
larryshamalama: This will be raised as a separate issue.

aeppl/mixture.py
```diff
@@ -248,14 +250,20 @@ def mixture_replace(fgraph, node):
...
         return None  # pragma: no cover
 
     old_mixture_rv = node.default_output()
 
+    if isinstance(node.op, Elemwise):
+        if not isinstance(node.op.scalar_op, Switch):
+            return None  # pragma: no cover
+        else:
+            idx, component_1, component_2 = node.inputs
+            node = (at.stack([component_1, component_2], axis=-1)[idx]).owner
```
larryshamalama, line 265: Are the `axis` always joined at `-1` for mixtures built with `switch`?

brandonwillard: I think that's what this will imply.

aeppl/mixture.py
```diff
@@ -238,7 +240,7 @@ def get_stack_mixture_vars(
     return mixture_rvs, join_axis
 
 
-@local_optimizer(subtensor_ops)
+@local_optimizer(subtensor_ops + (Elemwise,))
```
brandonwillard, line 243: A distinct `local_optimizer` is probably the best approach here.  

Also, if the `Op` class being tracked has only one type instance, that instance can be tracked directly, which makes the checking (i.e. `dict` hashing and `Op.__eq__`) of `Elemwise` `Op`s more accurate.  Specifically, the tracking functionality will use the `Elemwise.scalar_op` value, as well, so it won't need to be checked manually here.

aeppl/mixture.py
```diff
@@ -248,14 +250,20 @@ def mixture_replace(fgraph, node):
...
-
     rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
 
     if rv_map_feature is None:
         return None  # pragma: no cover
 
     old_mixture_rv = node.default_output()
 
+    if isinstance(node.op, Elemwise):
+        if not isinstance(node.op.scalar_op, Switch):
```
brandonwillard, line 261: A distinct `local_optimizer` would essentially move this check up the stack into `LocalOptGroup.transform`.

aeppl/mixture.py
brandonwillard, line 265: Looking at `get_stack_mixture_vars`, it seems like taking this approach (i.e. from within the `Subtensor`-parsing rewrite) will be a little too indirect (i.e. the `get_stack_mixture_vars` below is basically deconstructing the graph being constructed on this line).  Instead, most of the code from here to the end of the function can be refactored into its own helper function and called by two distinct local rewrites: one for `Subtensor`s and another for `Switch`es.

tests/test_mixture.py
```diff
@@ -705,3 +706,34 @@ def test_mixture_with_DiracDelta():
...
+
+    Z2_rv = at.stack(X_rv, Y_rv)[I_rv]
+    z2_vv = Z2_rv.clone()
+    z2_vv.name = "z2"
+
+    z1_logp = joint_logprob({Z1_rv: z1_vv, I_rv: i_vv}).eval({z1_vv: -10, i_vv: 0})
+    z2_logp = joint_logprob({Z2_rv: z2_vv, I_rv: i_vv}).eval({z2_vv: -10, i_vv: 0})
+
+    np.testing.assert_almost_equal(0.69049938, z1_logp)
+    np.testing.assert_almost_equal(0.69049938, z2_logp)
```
brandonwillard, line 739: For these symbolic things we should always try to directly assert the kinds of graphs we're expecting to get back.  Functions like `aesara.graph.basic.equal_computations` are usually the most direct way to do that; however, in this case, since we're dealing with a single node that holds all the relevant information, we can just inspect it directly (e.g. assert that the values of `fgraph.outputs[0].owner.inputs` are the expected `Variable`s&mdash;or their clones).

Numeric evaluation might only be helpful to confirm that the results of the underlying `switch`-to-`stack` conversion are correct, but I'm not sure that's worth testing in Aesara (e.g. it can be done directly in NumPy).  
Aside from that, a `logprob` test isn't entirely necessary, since it's testing the same mixture functionality that's already being tested above&mdash;and if it isn't, it should be tested above.  Regardless, it's not bad to have.

brandonwillard (changes requested):

aeppl/mixture.py
larryshamalama: What is this tracking functionality that you mention? I am not sure to fully understand the idea that you're conveying here. In the separate `local_optimizer` for `Elemwise`, I am still checking that the `scalar_op` is indeed a `Switch`. Is this still necessary?

brandonwillard: The first argument to `local_optimizer` is the list of `Op` instances or classes to which the rewrite is applied.  Having separate `local_optimizer`s for distinct `Op`s reduced the need for redundant/manual filtering within the rewrite.

Also, tracking something like `at.add` (i.e. an `Op` instance) will make sure that the rewrite sees only nodes using that `Op` instance, where as tracking `Elemwise` classes will require an `isinstance(op.scalar_op, aesara.scalar.basic.Add)` check within the rewrite.

aeppl/mixture.py
```diff
@@ -300,6 +302,50 @@ def mixture_replace(fgraph, node):
     return [new_mixture_rv]
 
 
+@local_optimizer((Elemwise,))
```
brandonwillard, line 305:
```suggestion
@local_optimizer((at.switch,))
```

Does that (and removing the `isinstance` below) work, as well?

larryshamalama: oh yes, it does! I didn't know that this was possible

tests/test_mixture.py
larryshamalama: I believe that #158 is a blocker for this PR, more specifically in writing a test to assess the equivalency between mixture graphs defined via `switch` and `stack`.

brandonwillard: > I believe that #158 is a blocker for this PR, more specifically in writing a test to assess the equivalency between mixture graphs defined via `switch` and `stack`.

That shouldn't be a _blocker_; we can always check the graphs manually.

brandonwillard (commented):
https://github.com/aesara-devs/aeppl/pull/163 should fix the issue you're seeing in CI.

larryshamalama:
Okay I see... I was wondering. I just re-polished `test_switch_mixture` in `test_mixture.py`

brandonwillard:
I've rebased and squashed this, so you'll need to pull your remote.

brandonwillard (commented):
Looks like this needs tests for the test value and name parts.

larryshamalama:
> Looks like this needs tests for the test value and name parts.

Thanks, I just added one in `test_compute_test_value`

PR #158, opened 2022-08-02
Add a custom metaclass for dynamic type creation of unmeasurable `Op`s

### LINKED ISSUE #157: Double `Unmeasurable` prefix in components of mixture sub-graphs
Mixture sub-graphs are identified by [mixture_replace](https://github.com/aesara-devs/aeppl/blob/main/aeppl/mixture.py#L241-L242) and each component is assigned to be unmeasurable [as per here](https://github.com/aesara-devs/aeppl/blob/main/aeppl/mixture.py#L274). However, in defining the stack, tensors which are combined go through a similar procedure [here](https://github.com/aesara-devs/aeppl/blob/main/aeppl/tensor.py#L175), hence why the type `UnmeasurableUnmeasurableNormalRV` occurs. This becomes an issue when trying to use functions such as [`aesara.graph.basic.equal_computations`](https://github.com/aesara-devs/aesara/blob/main/aesara/graph/basic.py#L1634) to check if Op types are identical.

Below is an example of this error:

```python
import aesara.tensor as at
from aeppl.opt import construct_ir_fgraph

srng = at.random.RandomStream(seed=2320)

I_rv = srng.bernoulli(0.5, name="I")
X_rv = srng.normal(-10., 0.1, name="X")
Y_rv = srng.normal(10., 0.1, name="Y")

# mixture from stack
Z_rv = at.stack([X_rv, Y_rv])[I_rv]
Z_rv.name = "Z"
z_vv = Z_rv.clone()

i_vv = I_rv.clone()

fgraph, _, _ = construct_ir_fgraph({Z_rv: z_vv, I_rv: i_vv})

fgraph.outputs[0] # Z-mixture
fgraph.outputs[0].owner.inputs # [NoneConst, ScalarFromTensor.0, X, Y]
type(fgraph.outputs[0].owner.inputs[2].owner.op) # aesara.tensor.random.basic.UnmeasurableUnmeasurableNormalRV
```

larryshamalama:
Closes #157.

This PR adds a check in `assign_custom_measurable_outputs` to verify if the Op name begins with the prefix and if the Op has already been assigned a dispatch `measurable_outputs_fn`.

However, one thing that I noticed is that nodes with the "same" `new_node_type` would go through the function again; in fact, I believe that only their hash values are not identical. Is this an issue? Below is an example of this.

```python
import aesara.tensor as at
from aeppl.opt import construct_ir_fgraph
from aeppl.abstract import _get_measurable_outputs

srng = at.random.RandomStream(seed=2320)

I_rv = srng.bernoulli(0.5, name="I")
X_rv = srng.normal(-10., 0.1, name="X")
Y_rv = srng.normal(10., 0.1, name="Y")

# mixture from stack
Z_rv = at.stack([X_rv, Y_rv])[I_rv]
Z_rv.name = "Z"
z_vv = Z_rv.clone()

i_vv = I_rv.clone()

registered_measurable_ops = list(_get_measurable_outputs.registry.keys())

# op1 and op2 are both UnmeasurableNormalRV
op1, op2 = registered_measurable_ops[5], registered_measurable_ops[6]

op1, op2 # (aesara.tensor.random.basic.UnmeasurableNormalRV,  aesara.tensor.random.basic.UnmeasurableNormalRV)

hash(op1), hash(op2) # (8793444441827, 8793444441922), not the same

hash(_get_measurable_outputs.registry[op1]), hash(_get_measurable_outputs.registry[op2]) # (373950558, 373950558), same hash values
```

Notice that `op1` and `op2` are no longer `UnmeasurableUnmeasurableNormalRV` with this PR.

CC @brandonwillard

### DIFF
--- a/aeppl/abstract.py
+++ b/aeppl/abstract.py
@@ -5,6 +5,7 @@ from typing import Callable, List
 
 from aesara.graph.basic import Apply, Variable
 from aesara.graph.op import Op
+from aesara.graph.utils import MetaType
 from aesara.tensor.random.op import RandomVariable
 
 
@@ -13,6 +14,29 @@ class MeasurableVariable(abc.ABC):
 
 
 MeasurableVariable.register(RandomVariable)
+
+
+class UnmeasurableMeta(MetaType):
+    def __new__(cls, name, bases, dict):
+        if "id_obj" not in dict:
+            dict["id_obj"] = None
+
+        return super().__new__(cls, name, bases, dict)
+
+    def __eq__(self, other):
+        if isinstance(other, UnmeasurableMeta):
+            return hash(self.id_obj) == hash(other.id_obj)
+        return False
+
+    def __hash__(self):
+        return hash(self.id_obj)
+
+
+class UnmeasurableVariable(metaclass=UnmeasurableMeta):
+    """
+    id_obj is an attribute, i.e. tuple of length two, of the unmeasurable class object.
+    e.g. id_obj = (NormalRV, noop_measurable_outputs_fn)
+    """
 
 
 def get_measurable_outputs(op: Op, node: Apply) -> List[Variable]:
@@ -70,7 +94,9 @@ def assign_custom_measurable_outputs(
     new_node = node.clone()
     op_type = type(new_node.op)
 
-    if op_type in _get_measurable_outputs.registry.keys():
+    if op_type in _get_measurable_outputs.registry.keys() and isinstance(
+        op_type, UnmeasurableMeta
+    ):
         if _get_measurable_outputs.registry[op_type] != measurable_outputs_fn:
             raise ValueError(
                 f"The type {op_type.__name__} with hash value {hash(op_type)} "
@@ -78,8 +104,11 @@ def assign_custom_measurable_outputs(
             )
         return node
 
+    new_op_dict = op_type.__dict__.copy()
+    new_op_dict["id_obj"] = (new_node.op, measurable_outputs_fn)
+
     new_op_type = type(
-        f"{type_prefix}{op_type.__name__}", (op_type,), op_type.__dict__.copy()
+        f"{type_prefix}{op_type.__name__}", (op_type, UnmeasurableVariable), new_op_dict
     )
     new_node.op = copy(new_node.op)
     new_node.op.__class__ = new_op_type
--- a/aeppl/abstract.py
+++ b/aeppl/abstract.py
@@ -69,15 +69,20 @@ def assign_custom_measurable_outputs(
 
     new_node = node.clone()
     op_type = type(new_node.op)
+
+    if op_type in _get_measurable_outputs.registry.keys():
+        if _get_measurable_outputs.registry[op_type] != measurable_outputs_fn:
+            raise ValueError(
+                f"The type {op_type.__name__} with hash value {hash(op_type)} "
+                "has already been dispatched a measurable outputs function."
+            )
+        return node
+
     new_op_type = type(
         f"{type_prefix}{op_type.__name__}", (op_type,), op_type.__dict__.copy()
     )
-
     new_node.op = copy(new_node.op)
     new_node.op.__class__ = new_op_type
-
-    # TODO: The above could be a stand-alone utility function for all sorts of
-    # instance-based dispatching
 
     _get_measurable_outputs.register(new_op_type)(measurable_outputs_fn)
 
--- a/tests/test_abstract.py
+++ b/tests/test_abstract.py
@@ -1,0 +1,94 @@
+import aesara.tensor as at
+import pytest
+from aesara.tensor.random.basic import NormalRV
+
+from aeppl.abstract import (
+    UnmeasurableVariable,
+    _get_measurable_outputs,
+    assign_custom_measurable_outputs,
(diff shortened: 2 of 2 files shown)


aeppl/abstract.py
```diff
@@ -69,15 +69,19 @@ def assign_custom_measurable_outputs(
...
-        f"{type_prefix}{op_type.__name__}", (op_type,), op_type.__dict__.copy()
-    )
-
-    new_node.op = copy(new_node.op)
-    new_node.op.__class__ = new_op_type
+    op_name = op_type.__name__
 
     # TODO: The above could be a stand-alone utility function for all sorts of
     # instance-based dispatching
+    if (
```
brandonwillard, line 76: We shouldn't use names for anything functional in Aesara-based code.

brandonwillard (changes requested):

aeppl/abstract.py
larryshamalama: How should we go about checking if an Op has already been dispatched a measurable_outputs_fn?

aeppl/abstract.py
```diff
@@ -69,16 +69,18 @@ def assign_custom_measurable_outputs(
 
     new_node = node.clone()
     op_type = type(new_node.op)
+
+    # TODO: The above could be a stand-alone utility function for all sorts of
+    # instance-based dispatching
```
larryshamalama, line 74: What does this TODO comment mean?

larryshamalama:
A test is failing with error `TypeError: __class__ assignment: 'UnmeasurableNormalRV' object layout differs from 'NormalRV'`: https://github.com/aesara-devs/aeppl/runs/7717160032?check_suite_focus=true

aeppl/abstract.py
```diff
@@ -15,6 +16,26 @@ class MeasurableVariable(abc.ABC):
 MeasurableVariable.register(RandomVariable)
 
 
+class UnmeasurableMeta(MetaType):
+    def __eq__(self, other):
+        if hasattr(self, "id_obj") and hasattr(other, "id_obj"):
+            return hash(self.id_obj) == hash(other.id_obj)
```
brandonwillard, line 22: Can you do an `issubclass(other, UnmeasurableMeta)` check instead, or something like it?  That, as well as adding/requiring `id_obj` in a `__new__` implementation would make these `__eq__` and `__hash__` implementations more direct, especially since we don't want `x == y` to be true for any `x` and `y` that have equal `id_obj` attributes.

aeppl/abstract.py
brandonwillard: You can remove this; it's an old TODO that no longer applies.

aeppl/abstract.py
```diff
@@ -69,16 +90,28 @@ def assign_custom_measurable_outputs(
 
     new_node = node.clone()
     op_type = type(new_node.op)
+
+    # TODO: The above could be a stand-alone utility function for all sorts of
+    # instance-based dispatching
+    if op_type in _get_measurable_outputs.registry.keys() and hasattr(
+        op_type, "id_obj"
+    ):
```
brandonwillard, line 98: Same here; can we use a `isinstance` or `issubclass` instead of `hasattr`?

larryshamalama:
Thanks @brandonwillard for the review. I will incorporate your suggestions shortly.

aeppl/abstract.py
larryshamalama: Absolutely. The correct check was `isinstance(other, UnmeasurableMeta)`

aeppl/abstract.py
```diff
@@ -15,6 +16,31 @@ class MeasurableVariable(abc.ABC):
 MeasurableVariable.register(RandomVariable)
 
 
+class UnmeasurableMeta(MetaType):
+    def __new__(cls, name, bases, dict):
+        if "id_obj" not in dict:
+            dict["id_obj"] = None
```
larryshamalama, line 22: Here, a check to ensure that id_obj is initialized was added. However, does this make line 41 `id_obj = None` redundant?

larryshamalama: To my own question: I believe so...

tests/test_abstract.py
```diff
@@ -0,0 +1,103 @@
...
+    unmeasurable_X2 = fgraph.outputs[0].owner.inputs[2].owner.op
+    unmeasurable_Y2 = fgraph.outputs[0].owner.inputs[3].owner.op
+
+    assert_equal_hash(unmeasurable_X2.__class__, unmeasurable_Y2.__class__)
+    assert_equal_hash(unmeasurable_X.__class__, unmeasurable_X2.__class__)
+
+    with pytest.raises(ValueError):
+        assign_custom_measurable_outputs(
+            fgraph.outputs[0].owner.inputs[2].owner, lambda x: x
+        )
```
brandonwillard, line 103: Since this is testing mixture-specific functionality, it should go with the mixture tests; however, I don't think we need this kind of test in the first place, because the branches checked by this test should already be covered by the existing mixture tests.

larryshamalama: I see. I just forced push replacing this mixture test with one that verifies that passing an unmeasurable node into `assign_custom_measurable_outputs` does not do anything.

brandonwillard (approved):

PR #161, opened 2022-08-17
Change use of "opt[imize]" to "rewrite"

brandonwillard:
This PR updates our use of "opt[imize]" to "rewrite".

PR #163, opened 2022-08-21
Fix Weibull parameterization

brandonwillard:
This PR fixes the parameterizations used for the Weibull distribution, which was changed to the NumPy parameterization in Aesara.


brandonwillard:
Looks like CI is picking up the 2.8.0 version from PyPI, so we might as well merge these fixes into #161.

PR #167, opened 2022-08-22
Set tests to error on uncaught warnings

### LINKED ISSUE #166: Set `filterwarnings` to `error` in the tests
Our tests should not allow un-caught/checked warnings.

The addition to `setup.cfg` mentioned [here](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html) should be sufficient.  We'll need to update our tests to prevent them from failing, though.

brandonwillard:
Closes #166


rlouf:
I rebased this PR on `main`, the tests fail because of the following warning:

```python
UserWarning: Rewrite warning: The Op i0 does not provide a C implementation. As well as being potentially slow, this also disables loop fusion.
```

`i0` seems to be the modified Bessel function of the first kind of order 0 (https://github.com/aesara-devs/aesara/blob/a1739f6c56f018d3c5508cd67e0e9a35617f66d6/aesara/tensor/math.py#L1402). Do we want a C implementation for this in Aesara or should we just ignore the warning?

brandonwillard:
> Do we want a C implementation for this in Aesara or should we just ignore the warning?

Ideally, we would have a C implementation, but we should simply catch/ignore acceptable warnings like this using `pytest.warns` or something similar.

PR #176, opened 2022-09-07
Implement logprob rewrites for rounding operations

ricardoV94:
The following graphs are now supported

```python
import aesara.tensor as at

x = at.random.normal()
xr = at.round(x)  # or at.floor/ at.ceil

xr_vv = xr.clone()
joint_logprob({xr: xr_vv})
```


ricardoV94:
CC @aseyboldt

aeppl/censoring.py
```diff
@@ -0,0 +1,214 @@
...
+from aesara.tensor.elemwise import Elemwise
+from aesara.tensor.var import TensorConstant
+
+from aeppl.abstract import MeasurableVariable, assign_custom_measurable_outputs
+from aeppl.logprob import CheckParameterValue, _logcdf, _logprob, logdiffexp
+from aeppl.rewriting import measurable_ir_rewrites_db
+
+
+class MeasurableClip(Elemwise):
+    """A placeholder used to specify a log-likelihood for a clipped RV sub-graph."""
```
brandonwillard, line 19: N.B. It would be much better if we could assign a custom IR class/type for the combination of `Elemwise` and `Clip` and not just the `Elemwise`.  Perhaps we could use a general `MeasurableElemwise` instead, because the name `MeasurableClip` is misleading when `Clip` is not involved at the type/class-level, nor is use of a scalar `Clip` `Op` enforced at run-time.

aeppl/censoring.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+MeasurableVariable.register(MeasurableClip)
+
+
+@node_rewriter(tracks=[Elemwise])
+def find_measurable_clips(
+    fgraph: FunctionGraph, node: Node
+) -> Optional[List[MeasurableClip]]:
+    # TODO: Canonicalize x[x>ub] = ub -> clip(x, x, ub)
```
brandonwillard, line 29: Do/will we have an issue for this?

brandonwillard (approved):

rlouf:
I see you also implemented logprob rewrites for `clip` as well. I've had to implement this by hand in the past, it's a nice addition. 

I would have maybe added a brief  explanation (with what this does mathematically speaking) in the docstring.

aeppl/censoring.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+MeasurableVariable.register(MeasurableRound)
+
+
+@node_rewriter(tracks=[Elemwise])
+def find_measurable_roundings(
+    fgraph: FunctionGraph, node: Node
+) -> Optional[List[MeasurableRound]]:
+    # TODO: Canonicalize x[x>ub] = ub -> clip(x, x, ub)
```
ricardoV94, line 140:
```suggestion
```

aeppl/censoring.py
```diff
@@ -0,0 +1,214 @@
...
+        value_upper = value + 0.5
+        value_lower = value - 0.5
+    elif isinstance(op.scalar_op, Floor):
+        value = at.floor(value)
+        value_upper = value + 1.0
+        value_lower = value
+    elif isinstance(op.scalar_op, Ceil):
+        value = at.ceil(value)
+        value_upper = value
+        value_lower = value - 1.0
```
ricardoV94, line 200: I am not sure about rounding the values. In theory they should be rounded already...

aeppl/censoring.py
ricardoV94: I'll add one

ricardoV94:
> I see you also implemented logprob rewrites for `clip` as well. I've had to implement this by hand in the past, it's a nice addition.

That was already implemented before

> I would have maybe added a brief explanation (with what this does mathematically speaking) in the docstring.

Where do you think such docstring would be helpful? On the logprob function?

aeppl/censoring.py
ricardoV94: Added a runtime check in the `clip_logprob`

aeppl/censoring.py
ricardoV94: https://github.com/aesara-devs/aeppl/issues/177

aeppl/censoring.py
brandonwillard: A run-time check can't really address this kind of type-level ambiguity.  Regardless, a sufficient run-time check would need to take place in this type's constructor, so that `MeasurableClip`s that don't use a scalar `Clip` `Op` can't be constructed.

ricardoV94: Added a` MeasurableElemwise` that checks for compatible scalar_op types at creation. What do you think?

PR #179, opened 2022-09-12
Raise RuntimeError when requested logprob expressions are not inferred

### LINKED ISSUE #90: Raise NotImplementedError when factorized_joint_logprob cannot retrieve terms for every value variable

ricardoV94:
When some logprob graphs cannot be derived, `factorized_joint_logprob` simply does not include them in the returned dictionary. This can be confusing, specially because cannot be expected to know what derivations aeppl is capable of doing in advance. This PR suggests raising explicitly whenever a requested logprob expression is missing by the end. Closes #90 

Before:

```python
import aesara.tensor as at

x = at.random.normal(name="x")
y = at.cos(at.random.normal())
y.name = "y"

x_vv = x.clone()
y_vv = y.clone()

# Aeppl does not know the logprob for y
print(factorized_joint_logprob({x: x_vv, y: y_vv}))  # {x: x_logprob}
```

After:

```python
import aesara.tensor as at

x = at.random.normal(name="x")
y = at.cos(at.random.normal())
y.name = "y"

x_vv = x.clone()
y_vv = y.clone()

# Aeppl does not know how to derive the logprob for y
# RuntimeError: The logprob terms of the following value variables could not be derived: {y}
factorized_joint_logprob({x: x_vv, y: y_vv})
```

There is a `UserWarning` that is issued by default when the normal variable in the graph of `y` is found, but it is not as informative to the user, and is not specific to failed derivations either.

### DIFF
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -175,6 +175,12 @@ def factorized_joint_logprob(
             for node in io_toposort(graph_inputs(q_logprob_vars), q_logprob_vars):
                 compute_test_value(node)
 
+    missing_value_terms = set(original_values.values()) - set(logprob_vars.keys())
+    if missing_value_terms:
+        raise RuntimeError(
+            f"The logprob terms of the following value variables could not be derived: {missing_value_terms}"
+        )
+
     return logprob_vars
 
 
--- a/tests/test_cumsum.py
+++ b/tests/test_cumsum.py
@@ -60,10 +60,8 @@ def test_destructive_cumsum_fails():
     """Test that a cumsum that mixes dimensions fails"""
     x_rv = at.random.normal(size=(2, 2, 2)).cumsum()
     x_vv = x_rv.clone()
-    with pytest.warns(UserWarning):
-        res = joint_logprob({x_rv: x_vv})
-
-    assert res is None
+    with pytest.raises(RuntimeError, match="could not be derived"):
+        joint_logprob({x_rv: x_vv})
 
 
 def test_deterministic_cumsum():
--- a/tests/test_mixture.py
+++ b/tests/test_mixture.py
@@ -40,8 +40,6 @@ def test_mixture_basics():
 
         return locals()
 
-    # TODO: This should raise explicitly, see #90
-    # with pytest.raises(ValueError, match=".*value variable was specified.*"):
     env = create_mix_model(None, 0)
     X_rv = env["X_rv"]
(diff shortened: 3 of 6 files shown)


rlouf:
Raising when the user intent is to get the logprob is a good idea. Maybe we could invite users to open an issue as well in the message?

ricardoV94:
> Raising when the user intent is to get the logprob is a good idea. Maybe we could invite users to open an issue as well in the message?

I would be reticent because we don't know if the failure is because of something we could derive but haven't implemented yet, or an invalid graph. That's why I opted for the `RuntimeError` instead of `NotImplementedError`

ricardoV94:
Yeah, some of the tests show that. Double counting is one such example:

https://github.com/aesara-devs/aeppl/blob/8fd78209940888d7ebdccd12cfa6e2572589237b/tests/test_truncation.py#L132-L155

rlouf (approved):

PR #180, opened 2022-09-14
Change joint_logprob sum default to False

### LINKED ISSUE #162: We should probably not sum `joint_logprob` by default
Currently, `joint_logprob` returns a scalar whatever the number of dimensions of the observation variable by default:

```python
import aesara.tensor as at
import numpy as np

a_rv = srng.normal(0, 1, size=10)

a_vv = a_rv.clone()
logprob = joint_logprob({a_rv: a_vv})

print(logprob.eval({a_vv: np.ones(10)}))
# -14.189385332046724
```

I think that this behavior is confusing. To understand why, consider two cases in which we&rsquo;d want the observation variable above to be a vector:

1.  If we have multiple (10) observations and would like to compute the joint probability of these observations as a function of the parameters, e.g. the likelihood.
2.  We want to sample from the posterior distribution with 10 chains.

In the first case, summation by default will work. In the second it will fail, and one will have to do instead:

```python
logprob = joint_logprob({a_rv: a_vv}, sum=False)
print(logprob.eval({a_vv: np.ones(10)}))
# [-1.41893853 -1.41893853 -1.41893853 -1.41893853 -1.41893853 -1.41893853  -1.41893853 -1.41893853 -1.41893853 -1.41893853]
```

I would rather have `joint_logprob` return a vector in this case and require the user to explicitly sum the result when computing the log-likelihood. This make *intent* more explicit in the code, and avoids the use of a keyword argument.

As an aside, I wouldn't have so much of a problem with the summation by default if we had a canonical way to handle multiple chains e.g. with something like `logprobs = elemwise(logprob)(np.ones(10))` (pseudo code). Maybe this exists and I am not aware of it?

ricardoV94:
Closes #162 

This PR brings `joint_logprop` and `factorized_joint_logprob` API very close to each other, so we may reconsider merging them.

### DIFF
--- a/aeppl/joint_logprob.py
+++ b/aeppl/joint_logprob.py
@@ -1,6 +1,6 @@
 import warnings
 from collections import deque
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Tuple, Union
 
 import aesara.tensor as at
 from aesara import config
@@ -73,7 +73,7 @@ def factorized_joint_logprob(
     Returns
     =======
     A ``dict`` that maps each value variable to the log-probability factor derived
-    from the respective `RandomVariable`.
+    from the respective `RandomVariable`, in the same order as the input `rv_values`
 
     """
     fgraph, rv_values, _ = construct_ir_fgraph(rv_values)
@@ -175,33 +175,39 @@ def factorized_joint_logprob(
             for node in io_toposort(graph_inputs(q_logprob_vars), q_logprob_vars):
                 compute_test_value(node)
 
-    missing_value_terms = set(original_values.values()) - set(logprob_vars.keys())
+    original_values_ordered = tuple(rv_values.values())
+
+    missing_value_terms = set(original_values_ordered) - set(logprob_vars.keys())
     if missing_value_terms:
         raise RuntimeError(
             f"The logprob terms of the following value variables could not be derived: {missing_value_terms}"
         )
 
-    return logprob_vars
+    # Returned logprob terms sorted by order of input `rv_values`
+    return {
+        original_value: logprob_vars[original_value]
+        for original_value in original_values_ordered
+    }
 
 
-def joint_logprob(*args, sum: bool = True, **kwargs) -> Optional[TensorVariable]:
+def joint_logprob(
+    *args, sum: bool = False, **kwargs
+) -> Union[TensorVariable, Tuple[TensorVariable, ...]]:
     """Create a graph representing the joint log-probability/measure of a graph.
 
     This function calls `factorized_joint_logprob` and returns the combined
-    log-probability factors as a single graph.
+    log-probability factors.
 
     Parameters
     ----------
-    sum: bool
-        If ``True`` each factor is collapsed to a scalar via ``sum`` before
-        being joined with the remaining factors. This may be necessary to
-        avoid incorrect broadcasting among independent factors.
+    sum: bool, default False
+        If ``False``, each logprob factor is returned as a separate output, in the same
+        order as the requested rv_values pairs. If ``True``, all the logprob factors
+        are summed reduced together into a scalar output.
 
     """
     logprob = factorized_joint_logprob(*args, **kwargs)
-    if not logprob:
-        return None
-    elif len(logprob) == 1:
+    if len(logprob) == 1:
         logprob = tuple(logprob.values())[0]
         if sum:
             return at.sum(logprob)
@@ -211,4 +217,4 @@ def joint_logprob(*args, sum: bool = True, **kwargs) -> Optional[TensorVariable]
         if sum:
             return at.sum([at.sum(factor) for factor in logprob.values()])
         else:
-            return at.add(*logprob.values())
+            return tuple(logprob.values())
--- a/aeppl/scan.py
+++ b/aeppl/scan.py
@@ -18,7 +18,7 @@ from aesara.tensor.var import TensorVariable
 from aesara.updates import OrderedUpdates
 
 from aeppl.abstract import MeasurableVariable, _get_measurable_outputs
-from aeppl.joint_logprob import factorized_joint_logprob
+from aeppl.joint_logprob import joint_logprob
 from aeppl.logprob import _logprob
 from aeppl.rewriting import (
     inc_subtensor_ops,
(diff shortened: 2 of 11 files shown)


aeppl/joint_logprob.py
```diff
@@ -175,33 +175,39 @@ def factorized_joint_logprob(
...
         raise RuntimeError(
             f"The logprob terms of the following value variables could not be derived: {missing_value_terms}"
         )
 
-    return logprob_vars
+    # Returned logprob terms sorted by order of input `rv_values`
+    return {
+        original_value: logprob_vars[original_value]
+        for original_value in original_values_ordered
+    }
```
ricardoV94, line 190: This can instead be done in `joint_logprob`, where it actually matters that the order of the returned graphs be predictable (if we decide to keep both functions)

brandonwillard (commented):
I'm cool with these changes, but I'm starting to question the name `joint_logprob` now.  Really, we need to map these functions to explicit mathematical expressions, and, in this case, what kind of expression does `joint_logprob` represent?

brandonwillard:
FYI: The current error should be fixed by the new Aesara release.

rlouf:
Now I think that the specific issue I brought up in #180 is resolved with support for gufuncs in Aesara.

The following does represent a model with 3 observations:

```
import aesara.tensor as at
import numpy as np

a_rv = srng.normal(0, 1, size=3)

a_vv = a_rv.clone()
logprob = joint_logprob({a_rv: a_vv})
```

While if I wanted to sample with 3 chains in parallel I should implement the following:

```
import aesara.tensor as at
import numpy as np

a_rv = srng.normal(0, 1)

a_vv = a_rv.clone()
logprob = joint_logprob({a_rv: a_vv})

logprob_vec = at.vectorize(logprob)
logprob_vec(at.as_tensor([1., 1., 1.]))
```

That's one thing. Now when it comes to naming, what we we usually call "logprob" corresponds in AePPL to `joint_logprob(sum=True)` and what we usually call "joint_logprob" corresponds "joint_logprob_factorized".

My suggestion here is to 
1. Rename  `joint_logprob_factorized` to `joint_logprob`;
2. Thus only keep the logic that corresponds to `sum=True` in `joint_logprob` that we rename to simply `logprob`.

As for my multiple chains thought experiment, until we have full support for gufuncs we would have to use the `size=num_chains` in the above example and then use `joint_logprob`. Which should be extra motivation to finish implementing the `Blockwise` Op.

brandonwillard:
> Now I think that the specific issue I brought up in #180 is resolved without ambiguity with support for gufuncs in Aesara.

Which issue?

rlouf:
Wrong one, sorry, I meant #162

ricardoV94:
> Now when it comes to naming, what we we usually call "logprob" corresponds in AePPL to joint_logprob(sum=True)

That's only true for the edge case of calling joint_logprob on a single variable. 

We have logprob as well, it's the dispatch function to obtain the logprob for a single measurable variables.

rlouf:
Let's forget about the current terminology in AePPL for a second, and focus on the objects we are manipulating and the definition of the quantities we want to get out of them.

Remember that what we call a model represents a distribution. The reason why we use probabilistic programming is because only a few of these distributions have a closed-form expression for their logprob, unlike named distributions like `Normal` and `Cauchy`.
But you can also see the named distribution as a "model". For instance forward sampling for the `Exponential` distribution can be implemented as:

```python
import aesara.tensor as at

lmbda_tt = at.scalar()
u_rv = srng.uniform()
e_tt = - at.log(u_rv) / lmbda_tt
```

And in the current setting you would use `joint_logprob(sum=True)` to get the equivalent of what `_logprob` gives you. So there's an inconsistency in the current naming.

I have spent a fair amount of time thinking about this terminology when working on `MCX` and what came out of this was, in terms of an interface that would be compatible with Aesara:
- `logprob({rv_1: vv_1, rv_2: vv_2})`, returns the total logprob of the model, i.e. what current corresponds to `joint_logprob(sum=True)`. In the toy example above you would want `logprob({u_rv: u_vv, e_tt: e_vv})` to return the same thing as `logprob(Exponential, lambda_tt)`, no?
- `joint_logprob({rv_1: vv_1, rv_2: vv_2})` returns the joint probability distribution on all possible pairs of output (that's what it is mathematically) i.e. what currently corresponds to `joint_logprob_factorized(({rv_1: vv_1, rv_2: vv_2})`

This is not an approach that's especially new and original. Take for instance the [PPL Oryx](https://www.tensorflow.org/probability/oryx/notebooks/probabilistic_programming#log_prob), which in Python is closest to Aesara in terms of general design.

ricardoV94:
> * `logprob({u_rv: u_vv, e_tt: e_vv})` to return the same thing as `logprob(Exponential, lambda_tt)`, no?

I guess you mean `logprob({e_tt: e_vv})` (you would only condition on one value, not two)? Yes, this is an edge case where one has to use `joint_logprob` to obtain the logprob of a single variable. We can certainly add a wrapper `logprob` that calls `factorized_joint_logprob` for a single value variable, is that what you mean?

Anything with more than one value variable produces joint logprobs (or mode logprob).

rlouf:
> We can certainly add a wrapper `logprob` that calls `factorized_joint_logprob` for a single value variable, is that what you mean?

That calls `joint_logprob(sum=True)`


> I would worry more about users calling logprob separately for each variable in a model, because 1) it's inefficient and 2) some logprobs can only be retrieved correctly with awareness of all variables in the model.

I'm not sure I understand what you're afraid people would do, can you give an example?

rlouf:
> The whole point of the joint_logprob wrapper right now is just to return non-dictionaries as outputs, I am also happy with removing it if it feels unnecessary. 

Then what prevents us from renaming `joint_logprob_factorized` to `joint_logprob` and `joint_logprob(sum=True)` to `logprob`?

ricardoV94:
> I'm not sure I understand what you're afraid people would do, can you give an example?

Simplest case is calling lopgrob of nested terms separately:

x = at.random.normal(0, name = "x")
y = at.random.normal(x, name="y")

x_vv = x.clone()
y_vv = y.clone()
logprob({x: x_vv}) + logprob({y: y_vv}) != logprob({x: x_vv, y_vv})
```

The graph of `logprob({y: y_vv})` will not depend on `x_vv` but `x`.

Others include double counting the same base variable. Or failing to account for transforms... Many ways for users to shoot themselves in the foot if they don't expect one rv graph -> one transformation -> one logp graph

rlouf:
Solved in #196, closing.

PR #196, opened 2022-10-31
Refactor the interface to construct logprob graphs

### LINKED ISSUE #162: We should probably not sum `joint_logprob` by default

rlouf:
Exploration following #193. Different approach from #180. Naming conventions follow Koller & Friedman "Probabilistic Graphical Models: Principles and Techniques".

We can now get a dictionary that maps the original random variables of a model to their associated conditional log-probabilities using:

```python
import aeppl

logprobs, values = aeppl.conditional_logprob(x_rv, y_rv, realized={z_rv: z_vv})
print(logprobs.keys())
# [x_rv, y_rv]
print(values)
# [x_vv, y_vv]
```

`conditional_logprob` replaces `factorized_joint_logprob`. I also removed the `sum` argument to `joint_logprob`; it is unclear to me what mathematical quantity `sum=False` corresponds to.

The choice of the general keyword _realize_ instead of _observed_ is intentional. The distinction shows up for instance when building the logprob for a MCMC step in a Metropolis-within-Gibbs sampler: one needs to condition on the values takes by the variables not targeted by the MCMC step, but we cannot really say that they're "observed". 

Closes #162


aeppl/joint_logprob.py
```diff
@@ -44,25 +44,34 @@ def factorized_joint_logprob(
...
 
-    If we create a value variable for ``Y_rv``, i.e. ``y_vv = at.scalar("y")``,
-    the graph of ``factorized_joint_logprob({Y_rv: y_vv})`` is equivalent to the
-    conditional probability :math:`\log p(Y = y \mid \sigma^2)`, with a stochastic
-    ``sigma2_rv``. If we specify a value variable for ``sigma2_rv``, i.e.
-    ``s_vv = at.scalar("s2")``, then ``factorized_joint_logprob({Y_rv: y_vv, sigma2_rv: s_vv})``
-    yields the joint log-probability of the two variables.
+    The graph generated by ``conditional_logprob(Y_rv)`` corresponds to the
+    conditional log-probability :math:`P(Y=y \mid \sigma^2)` where
+    :math:`\sigma^2` is a stochastic variables. The graphs generated by
```
brandonwillard, line 49:
```suggestion
    :math:`\sigma^2` is a stochastic variable. The graphs generated by
```

aeppl/joint_logprob.py
```diff
@@ -44,25 +44,34 @@ def factorized_joint_logprob(
...
+    .. code-block:: python
+
+        import aesara.tensor as at
+
+        sigma2_rv = at.random.invgamma(0.5, 0.5)
+        Y_rv = at.random.normal(0, at.sqrt(sigma2_rv))
+
+        logprobs = conditional_logprob(Y_rv, sigma2_rv)
+        # print(logprobs.keys())
+        # [sigma2_vv, Y_vv]
```
brandonwillard, line 67: Would we want the keys to be the generated value variables?  A caller will generally need to match the original random variable arguments with their respective results.  We could instead return a map from the original random variable arguments to their results.  We could also use the call argument order, though.

rlouf: It is indeed generally more convenient for callers to use random variables as keys in `logprobs`, I will make this change.

I also entertained the idea of implementing the following interface, where both maps are indexed by the random variables:

```python
# Random variables are passed as *args
logprobs, value_variables = conditional_logprob(Y_rv, sigma2_rv)
print(logprobs.keys())
# [Y_rv, sigma2_rv]

print(value_variables)
# {Y_rv: Y_vv, sigma2_rv: sigma2_vv}
```

This undoubtedly would be very convenient. However, there are cases in which we want to use value variables that contain data. The simplest solution would be to pass them as keyword arguments, but it does get clunky where also passing e.g. `extra_rewrites`:

```python
logprobs, value_variables = conditional_logprob(
    sigma2_rv,
    Y_rv=Y_data,
    extra_rewrites=rewrites
)
```

My opinion is that whatever is passed via `extra_rewrites` belongs to a separate function, but before (if) we do that we need to compose with it. Non-ambiguous alternatives:

```python
logprobs, value_variables = conditional_logprob(
    sigma2_rv,
    observed={Y_rv: Y_data},
    extra_rewrites=rewrites
)
print(value_variables)
# {sigma2_rv: sigma2_vv}

logprobs, value_variables = aeppl.conditional_logprob(sigma2_rv, Y_rv)
logprobs = aeppl.observe(logprobs, {Y_rv: Y_data})
```

But what is the difference between this and passing `Y_data` to the compiled function, really? I lean towards leaving "observation" considerations out of `AePPL` (it has a high-level wrapper feel to it) and go with `conditional_logprob(Y_rv, sigma2_rv)`. I really like the `conditional_logprob({x_rv: x_vv, Y_rv: Y_data})` interface but having to clone the random variables by hand quickly becomes tedious.

In short, there's no clear winner. We can readily eliminate the ambiguous option, but besides that it's all about the tradeoffs we're ready to make.

rlouf: Anyway the `observed=` interface is trivial enough to implement that we could add it to this PR.

rlouf:
I think we should also make it as easy as possible to compile a logprob graph, as long as the interface is as non-ambiguous and flexible as the current one. Let&rsquo;s consider a typical-sized model to illustrate the proposal:

```python
import aesara
import aesara.tensor as at

col_1 = at.as_tensor([1., 2., 3., 5., 1])
col_2 = at.as_tensor([1., 2., 3., 5., 1])

srng = at.random.RandomStream(3)

intercept_rv = srng.normal(-4, 1.)
beta_1_rv = srng.normal(0, 2.5)
beta_2_rv = srng.normal(0, 2.5)

likelihood = at.sigmoid(
    intercept_rv
    + beta_1_rv * col_1
    + beta_2_rv * col_2
)

logit_rv = srng.bernoulli(likelihood)
```

To compile the `joint_logprob` graph, one currently needs to do the following:

```python
import aeppl

intercept_vv = intercept_rv.clone()
beta_1_rv = beta_1_rv.clone()
beta_2_rv = beta_1_rv.clone()
logit_vv = logit_rv.clone()

logprob, values = aeppl.joint_logprob({
    intercept_rv: intercept_vv,
    beta_1_rv: beta_1_vv,
    beta_2_rv: beta_2_vv,
    logit_rv: logit_vv,
})

logrob_fn = aesara.function(values, logprob)
```

Having to create the value variables manually is cumbersome. We can easily improve this by making `joint_logprob` create the value variables and return them as a list. Compilation is now straightforward without any loss in generality:

```python
import aeppl

logprob, values = aeppl.joint_logprob(
    intercept_rv,
    beta_1_rv,
    beta_2_rv,
    logit_rv,
)

logrob_fn = aesara.function(values, logprob)
```

For `conditional_logprob` we need a bit more flexibility since we only need to specify the parents of a given variable when compiling its conditional log-probability, so I suggest to return the value variables in a map indexed by the corresponding random variables:

```python
import aeppl

logprobs, values = aeppl.conditional_logprob(
    intercept_rv,
    beta_1_rv,
    beta_2_rv,
    logit_rv,
)

logrob_intercept_fn = aesara.function([], logprobs[intercept_rv])
logrob_logit_fn = aesara.function(values.values(), logprobs[logit_rv])
```

We need to consider the scenario where the value that some RVs take is known at construction time. In this case, we know the values that `logit_rv` takes:

```python
position = at.as_tensor([0, 1, 1, 1, 1])

logprob, values = aeppl.joint_logprob(
    intercept_rv,
    beta_1_rv,
    beta_2_rv,
    realized={logit_rv: position},
)
```

This is good middle compromise between the explicit but cumbersome `conditional_logprob({sigma2_rv: sigma2_vv, Y_rv: Y_data})` and the restrictive `join_logprob(sigma2_rv, y_rv)` which requires to pass the data to the function at every iteration. The term `realized` is taken from measure theory, but we could also use the more pragmatic `observed`.

Anyway, these changes are fairly simple to implement and this could be done in this PR.

rlouf:
Now that I have implemented this interface in the last commit, I have a little experience using it given many tests needed to be updated. Here are my suggestions and remarks.

First, I think that the second ouput of `joint_logprob` and `conditional_logprob` should be a list in both cases. It is always easier to expand the tuple than to get elements from a dictionary:

### Always return a list of value variables

```python
logprob, (intercept_vv, beta_1_vv, beta_2_vv) = aeppl.joint_logprob(
    intercept_rv,
    beta_1_rv,
    beta_2_rv,
    realized={logit_rv: position},
)
```

This dictionary can always be built:

```python
rvs_to_sample = [intercept_rv, beta_1_rv, beta_2_rv]
logprob, values = aeppl.joint_logprob(*rvs_to_samples, realized={logit_rv: position})
rv_to_values = {rv: vv for rv, vv in zip(rvs_to_sample, values)}
```
but I expect this need to arise much less often than the need for convenience that the `list` output provides.

### Single random variable

I don't think we should special case the return values when a single random variable is passed to `joint_logprob` or `conditional_logprob`:

```python
logprob, (x_vv,) = joint_logprob(x_rv)

logprobs, (x_vv,) = conditional_logprob(x_rv)
print(logprobs)
# {x_rv: x_rv_logprob}
```

### The transforms interface is in the way

I can currently make the transforms interface work in a roundabout way by passing the corresponding random variables as "realized":

```python
  X_rv = at.random.halfnormal(0, 3, name="X")
  x_vv = X_rv.clone()
  x_vv.name = "x"

  transform_rewrite = TransformValuesRewrite({x_vv: DEFAULT_TRANSFORM})
  tr_logp, _ = joint_logprob(
      realized={X_rv: x_vv},
      extra_rewrites=transform_rewrite
)
```

I am not sure whether it is possible yet, but we could circumvent this by attributing transforms to the random variables (which makes more sense, it would even clarify this interface a little):

```python
X_rv = at.random.halfnormal(0, 3, name="X")

transform_rewrite = TransformValuesRewrite({x_rv: DEFAULT_TRANSFORM})
tr_logp, (x_vv,) = joint_logprob(X_rv,  extra_rewrites=transform_rewrite)
```

rlouf:
Current status:
- The interface described above (besides the transforms) has been implemented and the tests have been modified; I am quite happy with the result;
- I get an error related to a `test_value` that appear to not be computed, but I do not understand what change provoked this since they're all very benign. If someone well versed in test values want to take a look, I could do with the help
- I think I should move forward with the refactor of the `transform` interface described above.

brandonwillard:
> First, I think that the second ouput of `joint_logprob` and `conditional_logprob` should be a list in both cases. It is always easier to expand the tuple than to get elements from a dictionary:

Before I forget to mention this again: we could always return a `FunctionGraph` object that encapsulates the results.

brandonwillard:
> I don't think we should special case the return values when a single random variable is passed to `joint_logprob` or `conditional_logprob`:

Yeah, definitely not, that kind of stuff only ever complicates the typing and interfaces involved.

brandonwillard:
> I am not sure whether it is possible yet, but we could circumvent this by associating the transform with the random variables:

Random variables and their value variables should be one-to-one (most of the time, at least), no?

rlouf:
> Random variables and their value variables should be one-to-one (most of the time, at least), no?

You're right, it's more about how much time it will take than whether it is possible.

rlouf:
> Before I forget to mention this again: we could always return a `FunctionGraph` object that encapsulates the results.

What would be the benefit here ?

brandonwillard:
> > Before I forget to mention this again: we could always return a `FunctionGraph` object that encapsulates the results.
> 
> What would be the benefit here ?

It's an establish Aesara collection type that connects inputs and outputs, and, since we use them at different stages in the processes between all of the AePPL interface functions, it's worth considering an approach that uses them.  They can also hold meta information that a caller or the next interface function may want.

rlouf:
Just to be sure, are you suggesting we return?

```python
import aeppl

logprob = aeppl.joint_logprob(
    intercept_rv,
    beta_1_rv,
    beta_2_rv,
    logit_rv,
)

print(logprob.inputs)
# [intercept_vv, beta_1_vv, beta_2,_vv, logit_vv]
```

And 

```python
import aeppl

logprobs = aeppl.joint_logprob(
    intercept_rv,
    beta_1_rv,
    beta_2_rv,
    logit_rv,
)

print(logprobs)
# {
#    intercept_rv: intercept_logprob_fg,
#    beta_1_rv: beta_1_logprob_fg,
#    beta_2_rv: beta_2_logprob_fg,
#    logit_vv: logit_logprob_fg
# }
```

rlouf:
I have modified `joint_logprob` as to return a `FunctionGraph` locally. After using it I am not completely sold on the idea; to build the `FunctionGraph` we need to provide *all* the inputs to the logprob graphs. This undeniably makes compilation easier:

```python
import aesara
import aesara.tensor as at

import aeppl

M = at.matrix("M")
x = at.random.normal(0, 1, size=M.shape[1], name="X")
y = at.random.normal(M.dot(x), 1, name="Y")

logp = aeppl.joint_logprob(x, y)
print(logp.inputs)
# (M, X_vv, Y_vv)

logp_fn = aesara.function(logp.inputs, logp.ouputs)
```

But puts pressure on the internals of downstream users, such as AeMCMC, which only need the value variables. All that for a very relative gain over the more explicit:

```python
logp, value_variables = aeppl.joint_logprob(x, y)
print(value_variables)
# (X_vv, Y_vv)

logp_fn = aesara.function([M] + value_variables, logp)
```

Of course we could attach a `Feature` that holds the value variables and which could be used downstream callers. This adds a small cost, which would be outweighed if we can, down the line, need to target this region of the graph with logprob-specific rewrites. Still, I would hold off until that becomes useful/necessary.

brandonwillard (commented):
Why were the changes in "Remove dependency on specific error messages in tests" needed?  It's definitely preferable to have that kind of test specificity in these cases.

rlouf:
The tests pass, but I've noticed something while going over the tests that needs to be adressed:

```python
import aesara
import aesara.tensor as at


x_rv = srng.dirichlet(np.array([0.7, 0.3]))
transform_rewrite = TransformValuesRewrite({x_rv: DEFAULT_TRANSFORM})

# `test_transformed_logprob` passes if I pass this as `realized`
x_value_var = at.tensor(x_rv.dtype, shape=(None,) * x_rv.ndim)

# But it does not pass if I use the value variable produced here
logprob, (x_value_var,) = aeppl.joint_logprob(x_rv, extra_rewrites=transform_rewrite)
```

The solution seems to me to provide shape (and dtype) information to the value variables created inside `conditional_logprob`. Which would not be a bad thing, actually.

brandonwillard:
> The solution seems to me to provide shape (and dtype) information to the value variables created inside `conditional_logprob`. Which would not be a bad thing, actually.

What are the `TensorType`s in each case?

rlouf:
In the first case `TensorType(float64, (None, None))` and the second `TensorType(float64, (2, 2))`. The issue comes from the fact that the `x_value_var` passed in the current interface must be *in the transformed space*, thus of the same type as `transform.forward(rv.type())`, and the difference only shows up for `SimplexTransform`.

rlouf:
I fixed this by applying the forward transform to the copy of the corresponding random variable when initializing the value variables. This complicates the internals slightly but makes the interface less ambiguous.

If anything this shows that the changes proposed in #195 and #119  are necessary. The initialization would not be so complicated if, instead of passing a `TransformValuesRewrite` instance we could condition directly on the transformed variable as in the code below:

```python
import aeppl
import aesara.tensor as at

x_rv = at.random.normal(0, 1)
y = at.exp(x_rv)

logprob, (y_vv,) = aeppl.joint_logprob(y)
```

PR #208, opened 2022-12-02
Refactor `MeasurableElemwise` and its use

### LINKED ISSUE #170: Order of rewrites is only respected if kwarg `position` is used
I could be wrong, but when working on https://github.com/pymc-devs/pymc/pull/6072, it seemed that the order of the rewrites in the database is only respected when using the kwarg `position`. Otherwise it seems to be ignored, and only the order of definition matters.

So: 
https://github.com/aesara-devs/aeppl/blob/8ce3c63b39ef3e3019a88b7af95693d4a007ccb7/aeppl/rewriting.py#L234-L236

Should be:
```python
logprob_rewrites_db.register( 
     "measurable_ir_rewrites", measurable_ir_rewrites_db, "basic", position=-10, 
 )
```

brandonwillard:
This PR refactors the design of `MeasurableElemwise` and the rewrites that use it (i.e. the changes mostly introduced by https://github.com/aesara-devs/aeppl/pull/26).

These changes make it easier for others to extend the coverage of measurable `Elemwise` transforms using basic rewrite DB registration.  They also put each transformation rewrite under the management of Aesara's rewrite system independently, which means that they can be applied more efficiently (e.g. hash-based lookups on distinct `Op` instances) and their application can be enabled/disabled and tracked with more granularity&mdash;among other things.

The additions in #144 have also been added.

- [x] Consider replacing the negation and subtraction rewrites with direct transform-constructing rewrites to avoid breaking canonicalization (see the comments in #144)

Closes #170

### DIFF
--- a/aeppl/abstract.py
+++ b/aeppl/abstract.py
@@ -1,7 +1,7 @@
 import abc
 from copy import copy
 from functools import singledispatch
-from typing import Callable, List, Tuple
+from typing import Callable, List
 
 from aesara.graph.basic import Apply, Variable
 from aesara.graph.op import Op
@@ -121,16 +121,6 @@ def assign_custom_measurable_outputs(
 
 class MeasurableElemwise(Elemwise):
     """Base class for Measurable Elemwise variables"""
-
-    valid_scalar_types: Tuple[MetaType, ...] = ()
-
-    def __init__(self, scalar_op, *args, **kwargs):
-        if not isinstance(scalar_op, self.valid_scalar_types):
-            raise TypeError(
-                f"scalar_op {scalar_op} is not valid for class {self.__class__}. "
-                f"Acceptable types are {self.valid_scalar_types}"
-            )
-        super().__init__(scalar_op, *args, **kwargs)
 
 
 MeasurableVariable.register(MeasurableElemwise)
--- a/aeppl/censoring.py
+++ b/aeppl/censoring.py
@@ -1,13 +1,15 @@
-from typing import List, Optional
+from typing import TYPE_CHECKING, List, Optional
 
 import aesara.tensor as at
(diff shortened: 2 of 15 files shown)


rlouf:
This looks great, and now that it's cleaned up it looks more obvious to me what can be further improved.

Do you want to also go ahead and include the changes in #184 or do you want to leave it for future work?

### More general comment

I am not convinced that in the simple situations AePPL currently supports we need to rewrite the graph at all. That could be useful to have a "canonicalization" phase where we could rewrite transforms of `RandomVariable` when they return a known distribution (ratio of `StandardNormalRV` for instance). Here, the model graphs contain all the information we need. To be able to compute the model's logdensity without rewriting the graph we would need:

- To register every `RandomVariable`'s logdensity.
- To register the inverse transform of Aesara operators to compute the upstream value variables;
- To register the way transforms act on logdensities.

_However_, pushed a little further, rewriting the graph could be very interesting for AeMCMC. We would need to replace every random and observed variable in the graph by their corresponding measure, including their domain and logdensity graph. That means starting from the known measures and pushing them through the transforms, thus getting a graph of Measures; transforms are then irrelevant if we also keep track of the relation between value variables in parallel. This is probably what we should be aiming for in these rewrites, instead of replacing Aesara operators with `RVTransform`s

brandonwillard:
> Do you want to also go ahead and include the changes in #184 or do you want to leave it for future work?

Let's do that one in a follow-up.

> More general comment

Yes, there's a lot to rework here.  I'm going to merge this one so that&mdash;in the meantime&mdash;our existing functionality is preserved and better designed.  At the very least, it should be a lot easier to rewrite/redesign in this form compared to the current.

PR #210, opened 2022-12-05
Add support for power transforms

brandonwillard:
This PR adds support for general power transforms and replaces #184.  It uses a positive support in cases of ambiguity.


rlouf:
Just a note that #184 originally introduced a rewrite for `at.square` as well, but I think it would be better to only register `at.pow` here and make sure they're equivalent Aesara-side: https://github.com/aesara-devs/aesara/issues/1213

aeppl/transforms.py
```diff
@@ -579,17 +591,31 @@ def log_jac_det(self, value, *inputs):
...
+            at.eq(at.mod(power, 2), 0),
+            at.power(value, inv_power),
+            at.sgn(value) * at.power(at.abs(value), inv_power),
+        )
 
     def log_jac_det(self, value, *inputs):
-        return -2 * at.log(value)
+        from aeppl.logprob import xlogy0
+
+        power = self.transform_args_fn(*inputs)
```
rlouf, line 617: Don't you need to use `inv_power = at.reciprocal(power)` instead of `power` in the expression of the jacobian?

aeppl/transforms.py
```diff
@@ -422,8 +422,20 @@ def transform(measurable_input, *other_inputs):
 def measurable_reciprocal(fgraph, node):
     """Rewrite a `reciprocal` node to a `MeasurableVariable`."""
 
-    def transform(measurable_input, *other_inputs):
-        return ReciprocalTransform(), (measurable_input,)
+    new_node = at.power(node.inputs[0], at.as_tensor(-1)).owner
```
rlouf, line 425: What's the advantage of using `pow` to represent `reciprocal` here (besides a few less lines of code)? I feel such equivalences should be handled at the Aesara-level, either by aliasing or a canonicalization pass.

PR #221, opened 2023-01-07
Dynamically generate lists in documentation

brandonwillard:
This PR adds a custom `SphinxDirective` that dynamically generates cross-reference links to `RandomVariable`s in Aesara that are supported by `logprob` (i.e. that have a registered dispatch).

- [ ] Dynamically generate the `Op`s and `RVTransform`s in the "Invertible Transforms" section


rlouf:
What do the `TransformedX` distributions with `X` being a distribution registered with `_logprob` correspond to? In this case I think we should exclude them.

For distributions we also need a way to specify usage: the links refer to the `Op` documentation, but the way to instantiate these `Op`s (e.g. `at.random.normal`)  is nowhere specified. Should we had code examples in the `Op`'s docstrings directly?

We will also find a way to document the use with `RandomStream`, and only this one here.

brandonwillard:
> What do the `TransformedX` distributions with `X` being a distribution registered with `_logprob` correspond to? In this case I think we should exclude them.

Yes, we definitely need to exclude those.  They come from all the module-level `create_default_transformed_rv_op` calls in `aeppl.transforms`, I believe.

brandonwillard:
> For distributions we also need a way to specify usage: the links refer to the `Op` documentation, but the way to instantiate these `Op`s (e.g. `at.random.normal`) is nowhere specified. Should we had code examples in the `Op`'s docstrings directly?

We could also show a single example that explains most/all other cases; otherwise, we could attempt to generate RST for complete examples, but that seems like a lot.

rlouf:
> We could also show a single example that explains most/all other cases; otherwise, we could attempt to generate RST for complete examples, but that seems like a lot.

It is a lot, but it doesn't need to be done right now.

rlouf:
I have added a directive to list the invertible transforms, but the links are currently not showing up in the documentation.

Issue #238, opened 2023-02-25
Add relations to simplify sum of normally-distributed `RandomVariable`s


rlouf:
We should add rewrites that we can use to simplify the following expressions for `Z_rv`, `B_rv` and `C_rv` when they are being conditioned on:

```python 
import aesara.tensor as at

srng = at.random.RandomStream(0)
x_rv = srng.normal(0, 1)
y_rv = srng.normal(0, 2)
Z_rv = x_rv + y_rv # Normal RV


a_rv = srng.normal(0, 1, size=(10,))

B_rv = at.sum(a_rv) # Normal RV
C_rv = at.cumsum(a_rv) # Vector of normal RVs
```

larryshamalama:
I can give this a shot. Perhaps I can start with working with `Z_rv` and see how the logic generalizes to `B_rv` and `C_rv`?

brandonwillard:
> I can give this a shot. Perhaps I can start with working with `Z_rv` and see how the logic generalizes to `B_rv` and `C_rv`?

That would be great, thanks!

PR #239, opened 2023-03-07
Add rewrite for sum of normal RVs

larryshamalama:
Closes 1/3 of #238, for now. As I'm opening this PR, the rewrite works for scalar-valued normal RVs:

```python
import aesara.tensor as at
from aeppl.rewriting import construct_ir_fgraph

import aesara

X_rv = srng.normal(1, 3)
Y_rv = srng.normal(2, 4)
Z_rv = X_rv + Y_rv

fgraph, _, _ = construct_ir_fgraph({Z_rv: Z_rv.copy()})

aesara.dprint(fgraph.outputs[0])
# (below is subject to change...)
# ValuedVariable [id A]
#  |normal_rv{0, (0, 0), floatX, False}.1 [id B]
#  | |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x12EB2D7E0>) [id C]
#  | |TensorConstant{[]} [id D]
#  | |TensorConstant{11} [id E]
#  | |Elemwise{add,no_inplace} [id F]
#  | | |TensorConstant{1} [id G]
#  | | |TensorConstant{2} [id H]
#  | |Elemwise{sqrt,no_inplace} [id I]
#  |   |Elemwise{add,no_inplace} [id J]
#  |     |Elemwise{pow,no_inplace} [id K]
#  |     | |TensorConstant{3} [id L]
#  |     | |TensorConstant{2} [id M]
#  |     |Elemwise{pow,no_inplace} [id N]
#  |       |TensorConstant{4} [id O]
#  |       |TensorConstant{2} [id P]
#  |normal_rv{0, (0, 0), floatX, False}.1 [id B]
```

I'm happy to receive comments and then address the list below. The main to-do is to extend to matrix-valued normal RVs.

Some questions:
- Checking independence in AePPL. I believe that this identity would hold for conditional independence. I would have to think about this.
- What to put for the `rng` argument in the `make_node`? For now, I left it as the `rng` of the first RV input, but this is incorrect.
- I added the rewrite as an `EquilibriumGraphRewriter`. Also not sure about this.
- I added this rewrite in a new file, `math_stat.py`. Would there be a better name for this file? Or should this be added to an existing file?

Happy to hear any comments!

### DIFF
--- a/aeppl/__init__.py
+++ b/aeppl/__init__.py
@@ -12,6 +12,7 @@ from aeppl.printing import latex_pprint, pprint
 # isort: off
 # Add rewrites to the DBs
 import aeppl.censoring
+import aeppl.convolutions
 import aeppl.cumsum
 import aeppl.mixture
 import aeppl.scan
--- a/aeppl/convolutions.py
+++ b/aeppl/convolutions.py
@@ -1,0 +1,59 @@
+import aesara.tensor as at
+from aesara.graph.rewriting.basic import node_rewriter
+from aesara.tensor.random.basic import NormalRV, normal
+
+from aeppl.rewriting import measurable_ir_rewrites_db
+
+
+@node_rewriter((at.sub, at.add))
+def add_independent_normals(fgraph, node):
+    """Replace a sum of two un-valued independent normal RVs with a single normal RV."""
+
+    if node.op == at.add:
+        sub = False
+    else:
+        sub = True
+
+    X_rv, Y_rv = node.inputs
+
+    if not (X_rv.owner and Y_rv.owner) or not (
+        # This also checks that the RVs are un-valued (i.e. they're not
+        # `ValuedVariable`s)
+        isinstance(X_rv.owner.op, NormalRV)
+        and isinstance(Y_rv.owner.op, NormalRV)
+    ):
+        return None
+
+    old_rv = node.outputs[0]
+
+    mu_x, sigma_x, mu_y, sigma_y, _ = at.broadcast_arrays(
+        *(X_rv.owner.inputs[-2:] + Y_rv.owner.inputs[-2:] + [old_rv])
+    )
+
+    new_rng = X_rv.owner.inputs[0].clone()
+
+    new_node = normal.make_node(
+        new_rng,
+        old_rv.shape,
+        old_rv.dtype,
+        mu_x + mu_y if not sub else mu_x - mu_y,
+        at.sqrt(sigma_x**2 + sigma_y**2),
+    )
+
+    fgraph.add_input(new_rng)
+
+    # new_rng must be updated with values of the RNGs output by `new_node
+    new_rng.default_update = new_node.outputs[0]
+    new_normal_rv = new_node.default_output()
+
+    if old_rv.name:
+        new_normal_rv.name = old_rv.name
+
+    return [new_normal_rv]
+
+
+measurable_ir_rewrites_db.register(
+    "add_independent_normals",
+    add_independent_normals,
+    "basic",
+)
--- a/tests/test_convolutions.py
+++ b/tests/test_convolutions.py
@@ -1,0 +1,119 @@
+import aesara.tensor as at
+import numpy as np
+import pytest
+from aesara.tensor.random.basic import NormalRV
+
+from aeppl.rewriting import construct_ir_fgraph
+from aeppl.transforms import MeasurableElemwiseTransform
+
+
+@pytest.mark.parametrize(
+    "mu_x, mu_y, sigma_x, sigma_y, x_shape, y_shape",
+    [
+        (
+            np.array([1, 10, 100]),
+            np.array(2),
+            np.array(0.03),
+            np.tile(0.04, 3),
+            (),
+            (),
+        ),
+        (
+            np.array([1, 10, 100]),
+            np.array(2),
+            np.array(0.03),
+            np.full((5, 1), 0.04),
+            (),
+            (5, 3),
+        ),
+        (
+            np.array([[1, 10, 100]]),
+            np.array([[0.2], [2], [20], [200], [2000]]),
+            np.array(0.03),
+            np.array(0.04),
+            (),
+            (),
+        ),
+        (
+            np.broadcast_to(np.array([1, 10, 100]), (5, 3)),
+            np.array([2, 20, 200]),
+            np.array(0.03),
+            np.array(0.04),
+            (2, 5, 3),
+            (),
+        ),
+        (
+            np.array([[1, 10, 100]]),
+            np.array([[0.2], [2], [20], [200], [2000]]),
+            np.array([[0.5], [5], [50], [500], [5000]]),
+            np.array([[0.4, 4, 40]]),
+            (2, 5, 3),
+            (),
+        ),
+        (
+            np.array(1),
+            np.array(2),
+            np.array(3),
+            np.array(4),
+            (5, 1),
+            (1,),
+        ),
+    ],
+)
+@pytest.mark.parametrize("sub", [False, True])
+def test_add_independent_normals(mu_x, mu_y, sigma_x, sigma_y, x_shape, y_shape, sub):
+    srng = at.random.RandomStream(29833)
+
+    X_rv = srng.normal(mu_x, sigma_x, size=x_shape)
+    X_rv.name = "X"
+
+    Y_rv = srng.normal(mu_y, sigma_y, size=y_shape)
+    Y_rv.name = "Y"
+
+    Z_rv = X_rv + Y_rv if not sub else X_rv - Y_rv
+    Z_rv.name = "Z"
+    z_vv = Z_rv.clone()
+
+    fgraph, _, _ = construct_ir_fgraph({Z_rv: z_vv})
+
+    (valued_var_out_node) = fgraph.outputs[0].owner
+    # The convolution should be applied, and not the transform
+    assert isinstance(valued_var_out_node.inputs[0].owner.op, NormalRV)
+
+    new_rv = fgraph.outputs[0].owner.inputs[0]
+
+    new_rv_mu = mu_x + mu_y if not sub else mu_x - mu_y
+    new_rv_sigma = np.sqrt(sigma_x**2 + sigma_y**2)
+
+    new_rv_shape = np.broadcast_shapes(
+        new_rv_mu.shape, new_rv_sigma.shape, x_shape, y_shape
+    )
+
+    new_rv_mu = np.broadcast_to(new_rv_mu, new_rv_shape)
+    new_rv_sigma = np.broadcast_to(new_rv_sigma, new_rv_shape)
+
+    assert isinstance(new_rv.owner.op, NormalRV)
+    assert np.allclose(new_rv.owner.inputs[3].eval(), new_rv_mu)
+    assert np.allclose(new_rv.owner.inputs[4].eval(), new_rv_sigma)
+    assert new_rv.name == "Z"
+
+
+def test_normal_add_input_valued():
+    """Test the case when one of the normal inputs to the add `Op` is a `ValuedVariable`."""
+    srng = at.random.RandomStream(0)
+
+    X_rv = srng.normal(1.0, name="X")
+    x_vv = X_rv.clone()
+    Y_rv = srng.normal(1.0, name="Y")
+    Z_rv = X_rv + Y_rv
+    Z_rv.name = "Z"
+    z_vv = Z_rv.clone()
+
+    fgraph, _, _ = construct_ir_fgraph({Z_rv: z_vv, X_rv: x_vv})
+
+    valued_var_out_node = fgraph.outputs[0].owner
+    # We should not expect the convolution to be applied; instead, the
+    # transform should be (for now)
+    assert isinstance(
+        valued_var_out_node.inputs[0].owner.op, MeasurableElemwiseTransform
+    )
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -675,17 +675,6 @@ def test_transformed_rv_and_value():
     )
 
 
-def test_loc_transform_multiple_rvs_fails1():
-    srng = at.random.RandomStream(0)
-
-    x_rv1 = srng.normal(name="x_rv1")
-    x_rv2 = srng.normal(name="x_rv2")
-    y_rv = x_rv1 + x_rv2
-
-    with pytest.raises(DensityNotFound):
-        joint_logprob(y_rv)
-
-
 def test_nested_loc_transform_multiple_rvs_fails2():
     srng = at.random.RandomStream(0)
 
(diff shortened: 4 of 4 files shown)


aeppl/math_stat.py
```diff
@@ -0,0 +1,62 @@
...
+    if intersection != []:
+        return None
+
+    old_rv = node.outputs[0]
+
+    mu_x, sigma_x = X_rv.owner.inputs[-2:]
+    mu_y, sigma_y = Y_rv.owner.inputs[-2:]
+
+    new_node = normal.make_node(
+        X_rv.owner.inputs[0],  # temporary rng?
```
brandonwillard, line 40: This is a tough one we've been trying to deal with more generally.  The real complications come from any "updates" that are associated with the RNG objects.

More specifically, RNG objects are usually `SharedVariable`s, and those can have `SharedVariable.default_update` attributes that hold onto other `Variable`s (i.e. graphs representing the new value of the `SharedVariable` after each call to a compiled Aesara function with this update).  In the case of RNG objects created by `RandomStream`s, those default updates are the RNG objects output after drawing a sample from a `RandomVariable` node.  In other words, the updates mechanism replaces a shared RNG objects with a copy of the RNG object after a sample has been drawn using it, so the Aesara updates mechanism is emulating an in-place update of the RNG (e.g. just as `rng.normal()` automatically updates the internal state of `rng` in NumPy).

When performing replacements like this, it's possible that `rng = X_rv.owner.inputs[0]` will have a `rng.default_update` containing the original `X_rv` graph, and, if someone attempted to `aesara.compile` the resulting `new_node`, Aesara would pick up that old graph from the re-used RNG's `default_update` attribute and add it to the compiled results.  We definitely don't want to have to sample some unrelated graphs just to update the RNG objects, especially when those update graphs aren't consistent with the underlying sampling process.

Anyway, I'm pointing this out because it's a general design issue and usability complication of which I'm trying to make more people aware&mdash;mostly so we can fix it/improve the usability.

In this exact case, we can probably just clone the RNG `SharedVariable`s and add our own `default_update`s to those (e.g. similar to how `RandomStream.gen` does).  Also, the graphs produced here are only supposed to be used as an intermediate representation for obtaining log-probabilities, so, as long as we don't expect people to actually compile and sample these graphs, the default updates shouldn't matter, and we can probably just clone the RNGs and remove their default updates altogether.

brandonwillard (commented):
This looks good!  

I've added some comments regarding the handling of RNG objects, but that's a more general concern that we're trying to address in all of our rewriting work, and not a statement about this specific approach/implementation.

aeppl/math_stat.py
```diff
@@ -0,0 +1,62 @@
...
+        return None
+
+    X_rv, Y_rv = node.inputs
+
+    if not (X_rv.owner and Y_rv.owner) or not (
+        isinstance(X_rv.owner.op, NormalRV) and isinstance(Y_rv.owner.op, NormalRV)
+    ):
+        return None
+
+    # check independence
```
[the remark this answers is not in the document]
larryshamalama: I'm thinking more of the mathematical independence: $X \perp Y$. I am not even sure how this would manifest in Aesara/AePPL... We would have to use `MvNormalRV` instead?

aeppl/math_stat.py
larryshamalama: Hmm, there's a lot of info here and I have the general picture. Thanks for the explanation.

When you say clone the RNG `SharedVariable`, do you mean:
```python
rng = X_rv.owner.inputs[0].clone()
rng.default_update = something
```
I'm confused because it feels like there are two RNGs here, one from `X_rv` and one from `Y_rv`... Or should I create a new RNG object akin to what's being done in `RandomStream.gen`?

tests/test_math_stat.py
```diff
@@ -0,0 +1,96 @@
...
+        (
+            np.array(1),
+            np.array(6),
+            (5, 1),
+            (1,),
+            np.full((5, 3), 7),
+        ),
+    ],
+)
+def test_add_independent_normals(mu_x, mu_y, x_shape, y_shape, new_rv_mu):
```
rlouf, line 76: Would it make sense to compute `new_rv_mu` from `mu_x`, `mu_y`, `x_shape` and `y_shape` directly?

tests/test_math_stat.py
```diff
@@ -0,0 +1,96 @@
...
+    srng = at.random.RandomStream(29833)
+
+    X_rv = srng.normal(mu_x, 0.03)
```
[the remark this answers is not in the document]
larryshamalama: ...yes, oops

tests/test_math_stat.py
```diff
@@ -0,0 +1,96 @@
...
+    Z_rv.name = "Z"
+    z_vv = Z_rv.clone()
+
+    fgraph, _, _ = construct_ir_fgraph({Z_rv: z_vv})
+
+    new_rv = fgraph.outputs[0].owner.inputs[0]
+
+    assert isinstance(new_rv.owner.op, NormalRV)
+    assert np.allclose(new_rv.owner.inputs[3].eval(), new_rv_mu)
+    assert np.allclose(new_rv.owner.inputs[4].eval(), 0.05)
```
rlouf, line 95: Maybe we should make sure broadcasting also works when `sigma` is not a scalar?

aeppl/__init__.py
```diff
@@ -13,6 +13,7 @@
 # Add rewrites to the DBs
 import aeppl.censoring
 import aeppl.cumsum
+import aeppl.math_stat
```
rlouf, line 16: `convolution` is more specific than `math_stat`

aeppl/math_stat.py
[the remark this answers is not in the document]
rlouf: Afaiu two variables are (conditionally) independent if they have no common random variable ancestor in the graph. Did I miss something?

tests/test_math_stat.py
larryshamalama: When I was writing these tests, I thought that testing it on one parameter was sufficient under the assumption that `srng.normal` handles broadcasting the `sigma` the same way it does for `mu`. Is this the case? Would it still be better to include `sigma` broadcasting tests?

tests/test_math_stat.py
larryshamalama: Ah, just via NumPy manipulations? I think that works

brandonwillard:
> * What to put for the `rng` argument in the `make_node`? For now, I left it as the `rng` of the first RV input, but this is incorrect.

That's a very important question!  Those inputs are usually `SharedVariable`s with non-`None` `SharedVariable.default_update` values.  More specifically, those values are the RNG objects output by a `RandomVariable` `Op`&mdash;i.e. they're the updated RNGs obtained after sampling a random variable.  

Altogether, the `SharedVariable.default_update` value of an RNG is a `Variable` representing a graph that draws samples, and it's intended to serve as a way of specifying something like an in-place update on the RNG state.

Here's an illustration:
```python
import aesara
import aesara.tensor as at


srng = at.random.RandomStream(23092)

aesara.dprint(X_rv, print_default_updates=True)

# normal_rv{0, (0, 0), floatX, False}.1 [id A] 'X'
#  |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F44C55C1120>) [id B] <- [id A]
#  |TensorConstant{[]} [id C]
#  |TensorConstant{11} [id D]
#  |TensorConstant{0} [id E]
#  |TensorConstant{1} [id F]
#
# Default updates:
#
# normal_rv{0, (0, 0), floatX, False}.0 [id A]
#  |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F44C55C1120>) [id B] <- [id A]
#  |TensorConstant{[]} [id C]
#  |TensorConstant{11} [id D]
#  |TensorConstant{0} [id E]
#  |TensorConstant{1} [id F]
```

Notice the `.1` at the end of the first `normal_rv` line; that indicates that the second output (i.e. the sample value) is returned from the `RandomVariable` node given by `id A`.  The first input to that node is a `SharedVariable` RNG object (i.e. `id B`), and its debug print line ends in `[id B] <- [id A]`, which indicates that the `SharedVariable` is updated by one of the outputs of `id A`.  The default update graph printed at the end shows that the first output (i.e. the `.0`) of the same `id A` node is the "update value".  

When this graph is `aesara.function`-compiled, the `SharedVariable.default_update` graph will be discovered and added as a new output to the entire compiled graph, then, when the compiled function is evaluated, the `Linker` that executes the code will replace the value in the `SharedVariable` with the value of that new update output that was added (i.e. the RNG value after sampling).

In pure Python, this whole situation is similar to the following:
```python
from copy import copy
import numpy as np


rng = np.random.default_rng(2309)


def draw_normal(rng):
    new_rng = copy(rng)
    res = new_rng.normal(0, 1)
    return new_rng, res


# Replace the old `rng` with the returned RNG
rng, res = draw_normal(rng)
```

Now, problems arise when a `SharedVariable` RNG with a non-`None` `.default_update` is used as an input to a _different_ `RandomVariable` `Op`, because then the old sampling graph will show up in the compiled results.

This is what happens when we re-use an RNG with default updates for another graph:
```python
# The RNG from `X`
rng_X = X_rv.owner.inputs[0]

# Manually create a `RandomVariable` with a specific RNG
Y_rv = at.random.gamma(0.5, 0.5, name="Y", rng=rng_X)

# Compile the function
Y_rv_fn = aesara.function([], Y_rv)

# View the compiled graph
aesara.dprint(Y_rv_fn)

# gamma_rv{0, (0, 0), floatX, True}.1 [id A] 'Y' 1
#  |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F44C55C1120>) [id B]
#  |TensorConstant{[]} [id C]
#  |TensorConstant{11} [id D]
#  |TensorConstant{0.5} [id E]
#  |TensorConstant{2.0} [id F]
# normal_rv{0, (0, 0), floatX, False}.0 [id G] 0
#  |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F44C55C1120>) [id B]
#  |TensorConstant{[]} [id C]
#  |TensorConstant{11} [id D]
#  |TensorConstant{0} [id H]
#  |TensorConstant{1} [id I]
```

As you can see, the compiled graph is sampling a gamma **and** normal variable, but the normal sampling is only done in order to update the RNG `SharedVariable` used by the gamma!


If we want to reuse existing RNG `SharedVariable` objects we need to get rid of the default updates on them.  Ultimately, this is a big design complication, because it adds "state" to `RandomVariable`s and&mdash;more specifically&mdash;ties RNG objects to specific graphs.  Some related issues/discussions: https://github.com/aesara-devs/aesara/issues/898, https://github.com/aesara-devs/aesara/issues/738, https://github.com/aesara-devs/aesara/issues/543, https://github.com/aesara-devs/aesara/issues/454, https://github.com/aesara-devs/aesara/discussions/1251.

In AePPL, we generally don't sample the IR we produce, so this isn't an immediate problem, but it could easily creep into graphs somewhere down the line and cause real issues.

As far as this rewrite is concerned, we're probably fine copying the `SharedVariable` and (re)specifying the `.default_update` graph.

aeppl/math_stat.py
brandonwillard: > When you say clone the RNG `SharedVariable`, do you mean:

Yeah, with one important difference, though:
```python
new_rng = X_rv.owner.inputs[0].clone()
# something more needs to be done to `new_rng...

new_node = normal.make_node(rng, *other_inputs)
new_rng.default_update = new_node.outputs[0]
```

In other words, it's the `new_rng` that needs to be updated with the values of the RNGs output by `new_node`.

> I'm also confused because it feels like there are two RNGs here, one from `X_rv` and one from `Y_rv`

Which RNG we use is really just a choice.  The only thing we really need to consider is the user-level seeding, and, as long as we choose an RNG from one of the existing `RandomVariable`s, we should maintain some consistency with the seeding.  If we use `RandomStream.gen`, then we can't generate a connection with the user's seeding unless we have their `RandomStream` instance.

brandonwillard: Also, because this whole `RandomVariable` + `SharedVariable.default_update` situation is such a mess, I've created this: https://github.com/aesara-devs/aesara/issues/1478.

aeppl/convolutions.py
```diff
@@ -0,0 +1,56 @@
...
+    ):
+        return None
+
+    old_rv = node.outputs[0]
+
+    mu_x, sigma_x, mu_y, sigma_y, _ = at.broadcast_arrays(
+        *(X_rv.owner.inputs[-2:] + Y_rv.owner.inputs[-2:] + [old_rv])
+    )
+
+    new_rng = X_rv.owner.inputs[0]
```
larryshamalama, line 29: In one of @brandonwillard's comments in another discussion, we wrote
```python
new_rng = X_rv.owner.inputs[0].clone()
```
However, by cloning this `new_rng` for the `normal.make_node(...)` below, I obtain the following error:
```python
aesara.graph.utils.MissingInputError: Input 0 (RandomGeneratorSharedVariable(<Generator(PCG64) at 
0x7FF373115460>)) of the graph (indices start from 0), used to compute normal_rv{0, (0, 0), floatX, False}
(RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FF373115460>), TensorConstant{[5 1]}, 
TensorConstant{11}, Elemwise{add,no_inplace}.0, Elemwise{sqrt,no_inplace}.0), was not provided and not given a 
value. Use the Aesara flag exception_verbosity='high', for more information on this error.
```
I didn't spend that much time on this bug specifically, but I'm not sure why cloning the RNG yields a `MissingInputError`...

larryshamalama (commented):
Thanks @brandonwillard for the thorough reply. It took me a while to have an idea of the problem that you are addressing. @rlouf I incorporated your suggestions to the unit tests

> As far as this rewrite is concerned, we're probably fine copying the SharedVariable and (re)specifying the .default_update graph.

The only thing that I have yet to fully address is the RNG of the newly created Normal RV node. I gave some details as a comment in this code review

aeppl/convolutions.py
brandonwillard: Ah, yeah, you're seeing that error because `FunctionGraph` is really strict about adding new input variables (i.e. `Variable`s that have no/`None` `Variable.owner` values and are not `Constant`s) by default.  One quick way around this is to add the statement `fgraph.add_input(new_rng)`.

In the long run, we might want to make the `SharedVariable`s&mdash;or at least `RandomTypeSharedVariable`s&mdash;work like `Constant`s in this situation.  In general rewriting (e.g. for compilation purposes), that might not be the best idea, though, especially if there are `SharedVariable.default_update`s involved, because the associated updates won't be picked up during the regular `aesara.function` compilation process.

brandonwillard: You should be able to add `fgraph.add_input(new_rng)` to the rewrite, since `fgraph` is one of the arguments.

larryshamalama: Ah, of course... I just forced pushed again

larryshamalama:
@brandonwillard Is this PR close to the end? Should I get started on the remaining rewrites in #238 and build on top of this PR?

brandonwillard (commented):
> @brandonwillard Is this PR close to the end? Should I get started on the remaining rewrites in #238 and build on top of this PR?

As far as the rewrite is concerned, it looks good.  We can merge it as-is, ~but I don't think it will be used because of the rewrite ordering and existing transform-based approach.~

We'll need to change the rewrite DB setup in order to allow rewrites like this (i.e. ones that provide more/better distribution information) to come first in place of the generic transforms approach.  That concern is independent of this work, though.

brandonwillard:
OK, I'm about to push a change that adds support for subtraction and changes the DB to which the rewrite is registered.  I'll merge after that.

brandonwillard (approved):

brandonwillard:
This should be good to merge once the tests pass.

Thanks again, @larryshamalama; this was a great addition!

